PR #139, opened 2016-05-24
feat(enums): added new enums for some properties

dmiller15:
r?
@allisonheath @millerjs @tlicht3


allisonheath:
Can you verify that these enums cover all values currently stored?

dmiller15:
@allisonheath This is a bigger change than that. At the moment we are pulling ICD-O-3 values like c50.9 for primary_diangosis and site_of_resection_or_biopsy. We no longer want to do that. This needs to accompany changes in the mappings. I imagine none of the existing values will be maintained.

dmiller15:
These changes have been handled elsewhere in #184

PR #149, opened 2016-07-11
DAT-34 feat(demographic): unrequired year of birth

dmiller15:
r? @allisonheath 
https://jira.opensciencedatacloud.org/browse/DAT-34


allisonheath:
Do we want it to be preferred or no?

dmiller15:
Why make it preferred? Aren't we aiming to get rid of this entirely?

allisonheath:
Ah right, lgtm

PR #153, opened 2016-08-09
feat(terms): cleaned up and changed some CDEs

dmiller15:
r? @allisonheath @tlicht3 

- Cleaned up existing CDEs
- Moved some descriptions out of sample.yaml
- Added newly found CDEs for the percent_x properties

I imagine this will be changing quite a bit over the next few weeks but this is the best starting point to work from.


dmiller15:
Changes now covered in #161

PR #154, opened 2016-08-15
fix(rg): added HiSeq 4K instument model

dmiller15:
r? @allisonheath @msfitzsimons 

https://jira.opensciencedatacloud.org/browse/DM-12


allisonheath:
:+1:

PR #155, opened 2016-08-16
DAT-214 feat(sample): added non-required link to diagnosis

dmiller15:
r? @allisonheath @msfitzsimons @tlicht3 

Anyone have any better ideas for the "label"?

https://jira.opensciencedatacloud.org/browse/DM-13

### DIFF
--- a/gdcdictionary/schemas/sample.yaml
+++ b/gdcdictionary/schemas/sample.yaml
@@ -46,6 +46,12 @@ links:
     target_type: tissue_source_site
     multiplicity: one_to_one
     required: false
+  - name: diagnoses
+    backref: samples
+    label: related_to
+    target_type: diagnosis
+    multiplicity: many_to_one
+    required: false
 
 # Sample properties
 # Currently there are many TCGA-specific fields in the DB, this is an attempt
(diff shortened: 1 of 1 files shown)


tlicht3:
your'e asking about "diagnoses"? i like it... you're updating what is currently "diagnosis" to allow for more than one diagnosis per case, right?

dmiller15:
No I'm asking about the `label` field in the link. So right now I have `label: related_to`. This field is just for the semantic relation between the nodes. We would say `x sample is related_to y diagnosis`. The field can be anything really. I was just wondering if there was a better word for the relationship: like "derived_from, matched_to, taken_during..."

tlicht3:
oh sorry - it's not really derived from or taken during. the sample was classified with a diagnosis or assigned a diagnosis. related to is good, but are we going to give the a metastatic or recurrent tumor a different diagnosis? if so, related to might be a little broad because it's possible that the diagnosis will be slightly different for a met/recurrent tumor, but they're definitely related to the primary, but that's being awfully specific/literal.  when you create a label do you try to use something that could be used again or can it be specific to the relationship you're describing?

dmiller15:
I would say we aim to be specific with the label. Most of the labels we've come up with so far have been reused, but that has been more the result of not being able to come up with something better.

PR #158, opened 2016-08-23
DAT-108 feat(exposure): added new property and term for pack years smoked

dmiller15:
r? @allisonheath @msfitzsimons @sergeymk @tlicht3


sergeymk:
+1

Discussion here: https://jira.opensciencedatacloud.org/browse/DAT-108

millerjs:
Given that `cigarettes_per_day` is supposed to remain (though we will be nulling it out for TCGA exposure data), +1

PR #162, opened 2016-09-13
DAT-170:feat(new ocg analytes)

dmiller15:
r? @allisonheath @dankolbman 

https://jira.opensciencedatacloud.org/browse/DAT-170


allisonheath:
@dmiller15 can you include the ticket ID as well?

allisonheath:
Even with dictionary changes, in addition to a code review, we should expect a :white_check_mark: or other acknowledgement from QA that the PR can be merged.

That being said, even though this merge happened too quickly, reviewed with @BedfordWest and everything looks OK so we'll leave it in.

PR #168, opened 2016-09-21
chore(annotations): submittable <- true

millerjs:
Allow annotations to be submitted via the submission API.

Before we merge: _should_ they be submittable?

r? @dmiller15 @allisonheath @NCI-GDC/ucdevs


dmiller15:
Code looks fine. I'm a little dubious about them being submittable. How aware are the users of this node? Are they already submitting these?

dmiller15:
General consensus seems to be: no we should keep this not submittable.

PR #169, opened 2016-09-22
DAT-234 feat(mirna): added miRBase version property

dmiller15:
r? @allisonheath 

Minor issue for the time being as all our runs are miRBase 21. Not sure if there's another way of doing this, but here it is.

https://jira.opensciencedatacloud.org/browse/DAT-234


madopal:
Per Michael Fitz, we don't need this on the node any longer.

PR #171, opened 2016-09-29
DAT-223 feat(node): new node to capture biomarker information

dmiller15:
r? @allisonheath @tlicht3 @msfitzsimons @dankolbman 

https://jira.opensciencedatacloud.org/browse/DAT-223
Only the properties for BLGSP are in for the time being.

### DIFF
--- a/gdcdictionary/schemas/_terms.yaml
+++ b/gdcdictionary/schemas/_terms.yaml
@@ -281,6 +281,36 @@ base_caller_version: # TOREVIEW
   description: >
     Version of the base caller.
 
+biomarker_name:
+  description: >
+    The name of the biomarker being tested for this specimen and set of test results.
+  termDef:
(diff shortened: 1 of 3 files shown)


gdcdictionary/schemas/_terms.yaml
```diff
@@ -231,6 +303,50 @@ fastq_name: # TOREVIEW
   description: >
     Names of FASTQs.
 
+fev1_ref_post_bronch_%:
```
allisonheath: I'm not a huge fan of having special characters in property names, anyway we could avoid the `%`?

dmiller15: yes we can get rid of that.

dmiller15:
@allisonheath is this good to go now?

PR #173, opened 2016-10-27
DAT-295 feat(case): moved properties from project

dmiller15:
r? @allisonheath @dankolbman @tlicht3


dankolbman:
Going to play it safe and separate this into to commits.

PR #176, opened 2016-10-31
DAT-295/627 Remove project properties

dankolbman:
Removes `disease_type` and `primary_site` from project.
r? @dmiller15


dmiller15:
lgtm I take it we're closing #173?

shanewilson:
I thought we were leaving it is both places but making the project fields lists

@junjun-zhang

madopal:
We may eventually do this, but I think we're going to have to redo the PR anyway, so I'm closing for now.

Issue #179, opened 2016-11-14
Typo in file.yaml


bwalsh:
https://github.com/NCI-GDC/gdcdictionary/blob/0504d4544c8a973f0dd807eb305170232b5c1360/gdcdictionary/schemas/file.yaml#L82

should be `member_of`

```
  - name: tags
    backref: files
    label: memeber_of  
    target_type: tag
    multiplicity: many_to_many
    required: false
```

dankolbman:
Good find! I've submitted a fix for this.
Much appreciated!

bwalsh:
thanks!

majensen:
March 2016: https://jira.opensciencedatacloud.org/browse/SV-62
Also needs fixing in the data. One day....

purplepeopleeater:
Github issues are not routinely monitored by the GDC team. For questions to the GDC please contact the GDC Help Desk (support@nci-gdc.datacommons.io).
--GDC Support Team

PR #180, opened 2016-11-14
fix(typo): Fix type in file tag relation

dankolbman:
Addresses #179 
r? @dmiller15


dmiller15:
Sorry to dash your hopes; this is a known issue and has been addressed here: #88 and https://jira.opensciencedatacloud.org/browse/PGDC-1931

PR #185, opened 2016-12-15
feat/cleanup blgsp cptac: cleanup new properties

dmiller15:
r? @allisonheath @millerjs @dankolbman @tlicht3 

Some fixes as recommended by Tara. Should go in to this data release.


gdcdictionary/schemas/diagnosis.yaml
```diff
@@ -525,44 +495,30 @@ properties:
...
-      $ref: "_terms.yaml#/hiv_status"
+      $ref: "_terms.yaml#/hiv_positive"
     enum:
-      - Equivocal
-      - Indeterminate
-      - Not Done
-      - Not Available
-      - Negative
-      - Positive
+      - "Yes"
```
millerjs: Is there a reason this is an enum, not a `boolean`? Do we expect more non-binary fields in the future?

gdcdictionary/schemas/diagnosis.yaml
```diff
@@ -610,32 +559,15 @@ properties:
   lymph_nodes_positive:
     term:
       $ref: "_terms.yaml#/lymph_nodes_positive"
-    enum:
-      - "Yes"
-      - "No"
-      - Unknown
-      - Not Reported
-      - Not Allowed To Collect
+    type: number
```
millerjs: Is `number` correct here?

gdcdictionary/schemas/diagnosis.yaml
dmiller15: We've had this type of yes/no/unknown structure in many previous fields. How would it look as a boolean? Would we be able to capture "Unknown"?

gdcdictionary/schemas/diagnosis.yaml
dmiller15: According to Tara, yes this is an actual count.

allisonheath: `number` or `integer` ?

gdcdictionary/schemas/diagnosis.yaml
millerjs: Didn't see `unkown` there, this could be captured with a provided `null`, though a boolean would prevent future intermediate values

PR #194, opened 2017-02-15
DAT-30: feat(demographic): remove year of birth

dmiller15:
remove year of death

r? @madopal


madopal:
Apparently, we're keeping this for now. So I'm closing.

PR #199, opened 2017-05-15
TT-75 feat(FMI): one genomic_file to many read_groups

dmiller15:
r?
https://jira.opensciencedatacloud.org/browse/TT-75


dmiller15:
We have decided to split the submitted genomic profile file into its individual pieces. As such this is no longer required.

PR #211, opened 2017-07-05
TT-92 feat(TT-92): added new node for somatic mutation indexes

dmiller15:


gdcdictionary/schemas/somatic_mutation_index.yaml
```diff
@@ -0,0 +1,76 @@
+$schema: "http://json-schema.org/draft-04/schema#"
+
+id: "somatic_mutation_index"
+title: Somatic Mutation Index
+type: object
+namespace: http://gdc.nci.nih.gov
```
ProfOak: Is there a reason we're using the old url?

dmiller15: not in particular. i can change it

PR #218, opened 2017-08-08
Feat/api 315 all caps new projects and programs

jesuspguofc:
Renamed branch.

### DIFF
--- a/gdcdictionary/schemas/program.yaml
+++ b/gdcdictionary/schemas/program.yaml
@@ -36,6 +36,7 @@ properties:
   name:
     type: string
     description: "Full name/title of the program."
+    pattern: "^[A-Z,0-9]+|[0-9]+|[A-Z]+$"
   dbgap_accession_number:
     type: string
     description: "The dbgap accession number provided for the program."
--- a/gdcdictionary/schemas/project.yaml
+++ b/gdcdictionary/schemas/project.yaml
@@ -67,6 +67,7 @@ properties:
     description: "Project code" # TOREVIEW
   disease_type:
     type: string
+    pattern: "^[A-Z,0-9]+|[0-9]+|[A-Z]+$"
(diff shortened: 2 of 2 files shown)


gdcdictionary/schemas/program.yaml
```diff
@@ -36,6 +36,7 @@ properties:
   name:
     type: string
     description: "Full name/title of the program."
+    pattern: "^[A-Z0-9]+|[0-9]+|[A-Z]+$"
```
philloooo: is this saying the name should start with uppercase alphabetical or number and end with only uppercase alphabetical?

jesuspguofc: Hi Phillis, it is saying that we can have project names with numbers or uppercase characters only, it is hard to see but there are pipes (or's) in between. Probably there is a better why of structuring this. I am open to suggestions :)
>>> import re
>>> pattern = "^[A-Z0-9]+|[0-9]+|[A-Z]+$"
>>> re.match(pattern, "A")
<_sre.SRE_Match object at 0x7efea15d9370>
>>> re.match(pattern, "9")
<_sre.SRE_Match object at 0x7efea15d93d8>
>>> re.match(pattern, "999")
<_sre.SRE_Match object at 0x7efea15d9370>
>>> re.match(pattern, "AAAA")
<_sre.SRE_Match object at 0x7efea15d93d8>
>>> re.match(pattern, "9AAAA")
<_sre.SRE_Match object at 0x7efea15d9370>
>>> re.match(pattern, "A9AAAA")
<_sre.SRE_Match object at 0x7efea15d93d8>

However I just found another bug, thanks for asking!

jesuspguofc: Just to clarify, I need to add the $ symbol to the other 2 options, to signify that no other trailing characters are allowed.

PR #221, opened 2017-08-15
Change GDCDictionary init to bomb out with unicode in schemata files

jbarno:
(cherry picked from commit 6dad7862181135a6db5fc3e558f67f7c75adcc79)

Verified this breaks everything if there's a Unicode char in the schemata (But not in terms or definition files)

### DIFF
--- a/gdcdictionary/python/__init__.py
+++ b/gdcdictionary/python/__init__.py
@@ -27,28 +27,6 @@ def visit_directory(path):
     finally:
         os.chdir(cdir)
 
-
-def load_yaml(name):
-    """Return contents of yaml file as dict"""
-    with open(name, 'r') as f:
-        return yaml.safe_load(f)
-
-
-def load_schemas_from_dir(directory):
-    """Returns all yamls and resolvers of those yamls from dir"""
-
-    schemas, resolvers = {}, {}
-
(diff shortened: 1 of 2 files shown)


jbarno:
Here's an example failing travis build: https://travis-ci.org/NCI-GDC/gdcdictionary/builds/264789972

gdcdictionary/python/__init__.py
```diff
@@ -79,10 +57,40 @@ def __init__(self, lazy=False, root_dir=None, definitions_paths=None,
...
+        with open(name, 'r') as f:
+            if name in self.exclude:
+                return yaml.safe_load(f)
+            else:
+                try:
+                    ascii_file = f.read().encode("ascii")
+                except UnicodeDecodeError:
+                    print "Error in file: {}".format(name)
+                    raise
+                return yaml.safe_load(ascii_file)
```
jbarno: Might be safer to just safe_load the original file instead?

jbarno:
Note the cherry-picking commit messages are from my branch `jbarno/test-DAT-1064-unicode-chars` which was my original unit-test try but that branch was gross. I add -x to my cherry-picks out of habit.

PR #246, opened 2017-12-13
TT-275: new sample types for HCMI

dmiller15:
https://jira.opensciencedatacloud.org/browse/TT-257


madopal:
BTW, this is actually TT-257, not TT-275

PR #307, opened 2018-03-08
TT-422: alignment link no longer exclusive

dmiller15:
feat(TT-422): updated example for nonexclusivity

https://jira.opensciencedatacloud.org/browse/TT-422


dmiller15:
Handled in #401

PR #308, opened 2018-03-08
TT-412: update prior malignancy description

dmiller15:
https://jira.opensciencedatacloud.org/browse/TT-412


dmiller15:
Handled in #400

PR #309, opened 2018-03-08
TT-403: file_size type integer

dmiller15:
https://jira.opensciencedatacloud.org/browse/TT-403


madopal:
Closing because I rebased develop, and this caused issues. Will reopen against release/horton.

PR #326, opened 2018-03-13
TT-379: Add two new fields to demographic node

wwysoc2:
Two new fields were added to the demographic node: premature_at_birth and weeks_gestation_at_birth


gdcdictionary/schemas/demographic.yaml
```diff
@@ -146,6 +146,20 @@ properties:
       - Unknown
       - Not Reported
 
+  premature_at_birth:
+    term:
+      $ref: "_terms.yaml#/premature_at_birth"
+    enum:
+      - Yes
+      - No
```
dmiller15: Yes and No need to be put in quotes. `"Yes"` and `"No"`

dmiller15:
The alcohol_days_per_week and alcohol_drinks_per_day fields are not supposed to be in this PR. Please remove.

PR #328, opened 2018-03-13
TT-442: Removed enum values from read_groups.library_strategy

wwysoc2:
https://jira.opensciencedatacloud.org/browse/TT-442

"Amplicon", "Validation" and "Other" enum values were removed from read_groups.library_strategy.


dmiller15:
alcohol_days_per_week and alcohol_drinks_per_day have been removed. All good now. 👍

PR #331, opened 2018-03-14
TT-379: Add two new fields to demographics

wwysoc2:
https://jira.opensciencedatacloud.org/browse/TT-379

Added two new fields, updated _terms, updated valid example JSON


dmiller15:
Needs to be a merge against release/horton.

wwysoc2:
Merging against release/horton now

PR #351, opened 2018-03-19
feat(TT-429):Added new property submission_disabled

ganjigajanan:
https://jira.opensciencedatacloud.org/browse/TT-429

Pls Review.

### DIFF
--- a/gdcdictionary/schemas/project.yaml
+++ b/gdcdictionary/schemas/project.yaml
@@ -141,6 +141,12 @@ properties:
     default: false
     type: boolean
 
+  submission_enabled:
+    description: >
+      Indicates if submission to a project is allowed.
+    default: true
+    type: boolean
+
   programs:
     description: >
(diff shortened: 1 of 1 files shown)


gdcdictionary/schemas/project.yaml
```diff
@@ -141,6 +141,12 @@ properties:
     default: false
     type: boolean
 
+  submission_disabled:
+    description: >
+      Indicates if submission to a project is allowed.
```
dmiller15: Well technically, this field will indicate if submission is disabled.

gdcdictionary/schemas/project.yaml
```diff
@@ -141,6 +141,12 @@ properties:
     default: false
     type: boolean
 
+  submission_enabled:
+    description: >
+      Indicates if submission to a project is allowed.
+    default: false
```
dmiller15: Default should be true now.

PR #362, opened 2018-03-26
TT-493: Added 261 new values to treatment.anatomic_site

wwysoc2:
https://jira.opensciencedatacloud.org/browse/TT-493

Added a larges list of enum values to treatment.treatment_anatomic_site. Added "Throat" to list and changed "&" to "and"


gdcdictionary/schemas/treatment.yaml
```diff
@@ -113,54 +212,217 @@ properties:
...
+      - Lymph Node(s) Parotid
+      - Lymph Node(s) Pelvic
+      - Lymph Node(s) Popliteal
+      - Lymph Node(s) Retroperitoneal
+      - Lymph Node(s) Splenic
+      - Lymph Node(s) Subclavicular
+      - Lymph Node(s) Submandibular
+      - Lymph Node(s) Supraclavicular
       - Lymph Nodes
+      - Lymph Nodes(s) Mediastinal
```
dmiller15: It's from the ticket. I take it this is an intentional misspelling?

wwysoc2: Good catch! All of the rest of the "Lymph Node(s)" seem to be okay. Unless this will de-synchronize us from some external list, I think it should be corrected.

PR #384, opened 2018-03-30
TT-525: rename masked somatic mutations yaml

dmiller15:
https://jira.opensciencedatacloud.org/browse/TT-525


dmiller15:
Not worth the effort to resolve the conflict. Will be dealt with in the final Horton cleanup PR.

PR #385, opened 2018-03-30
TT-501: added node property downloadable

dmiller15:
fix(TT-501): modified metaschema for new prop

fix(TT-501): more metaschema changes

https://jira.opensciencedatacloud.org/browse/TT-501


gdcdictionary/schemas/aligned_reads_index.yaml
```diff
@@ -10,6 +10,7 @@ project: '*'
 description: "Data file containing the index for a set of aligned reads."
 additionalProperties: false
 submittable: true
+downloadable: false
```
madopal: We messed up, ticket updated, but this & somatic_mutation_index need to be downloadable.

gdcdictionary/schemas/somatic_mutation_index.yaml
```diff
@@ -10,6 +10,7 @@ project: '*'
 description: "Data file containing the index for a set of aligned reads."
 additionalProperties: false
 submittable: true
+downloadable: false
```
madopal: This was the other one that should be true. Again, sorry, ticket updated.

PR #391, opened 2018-04-02
Feat/tt 417 remove treatment days2treatment rollback

madopal:
I think we still got a few changes in here, but the only data-related one I can see is yanking the colon from the enum, which is a Good Thing :tm:.


dmiller15:
Handled in #397

PR #392, opened 2018-04-02
Feat/tt 403 file size integer rollback

madopal:
Yet another fix, should have the same commits in as the others.


dmiller15:
Handled in #398

PR #419, opened 2018-05-08
fix(submitter_id): require submitter_id everywhere

DrSpez:
Add `submitter_id` to required fields list on all nodes


DrSpez:
change is implemented in #423

PR #421, opened 2018-05-10
TT-609: required submitter_id on all nodes

dmiller15:
https://jira.opensciencedatacloud.org/browse/TT-609


gdcdictionary/schemas/exposure.yaml
```diff
@@ -30,6 +30,7 @@ links:
     required: true
 
 preferred:
+  - submitter_id
```
jbarno: @madopal is this in the wrong section?

PR #426, opened 2018-06-26
TT-660: new rna expression workflow type

dmiller15:
https://jira.opensciencedatacloud.org/browse/TT-660


ganjigajanan:
Looks good.

PR #433, opened 2018-06-26
TT-683: updated days_to_birth/death descriptions

dmiller15:
https://jira.opensciencedatacloud.org/browse/TT-683


ganjigajanan:
Looks good. @dmiller15 , @madopal this needs to be merged in the end as there are other tickets which update the _term.yaml file

PR #440, opened 2018-07-05
DAT-1525: unknown/not reported to qc state props

dmiller15:
https://jira.opensciencedatacloud.org/browse/DAT-1525


ganjigajanan:
Looks good

PR #447, opened 2018-07-09
DAT-1554: updated experimental_strategy values

dmiller15:
https://jira.opensciencedatacloud.org/browse/DAT-1554


ganjigajanan:
Looks good. @dmiller15 , @madopal  this needs to be merged in the end as there are other tickets which add enums to other properties on this node.

PR #448, opened 2018-07-09
DAT-1556: updated simple germline experimental strat

dmiller15:
https://jira.opensciencedatacloud.org/browse/DAT-1556


ganjigajanan:
Looks good. @dmiller15 , @madopal  this needs to be merged in the end as there are other tickets which add enums to other properties on this node.

PR #452, opened 2018-07-09
TT-686: removed ldh_level_at_diagnosis property

dmiller15:
https://jira.opensciencedatacloud.org/browse/TT-686


ganjigajanan:
Looks   good. @dmiller15 , @madopal this needs to be merged in the end as there are other tickets which add enums to other properties on this node. possible collision with https://jira.opensciencedatacloud.org/browse/DAT-1560

PR #453, opened 2018-07-09
DAT-1519: aliquot exclusive link with sample/aliquot

dmiller15:
https://jira.opensciencedatacloud.org/browse/DAT-1519


ganjigajanan:
Looks good

PR #455, opened 2018-07-12
DAT-1579: new vascular envasion type enums

dmiller15:
https://jira.opensciencedatacloud.org/browse/DAT-1579


ganjigajanan:
Looks good

PR #469, opened 2018-07-16
DAT-1584: update disease response enum

dmiller15:
https://jira.opensciencedatacloud.org/browse/DAT-1584


ganjigajanan:
looks good

PR #479, opened 2018-07-23
DAT-1594: updated followup terms

dmiller15:
https://jira.opensciencedatacloud.org/browse/DAT-1594


ganjigajanan:
Looks good. @dmiller15 , @madopal  this needs to be merged in the end as there are other tickets which update the _term.yaml file

PR #492, opened 2018-08-09
DAT-1668: new aligned reads props

dmiller15:
https://jira.opensciencedatacloud.org/browse/DAT-1668


gdcdictionary/schemas/aligned_reads.yaml
```diff
@@ -73,6 +73,21 @@ uniqueKeys:
 properties:
   $ref: "_definitions.yaml#/data_file_properties"
 
+  average_base_quality:
+    term:
+      $ref: "_terms.yaml#/average_base_quality"
+    type: number
```
madopal: Is it ok that this is int or float? Or do we want to make it one or the other?

PR #552, opened 2018-11-12
Fix/experiment metadata

giangbui:


madopal:
This has either the wrong starting branch, or it's being merged to the wrong branch. We'll figure out where to merge it, but this isn't it.

PR #572, opened 2018-12-07
Fix/experiment metadata

giangbui:
Add required link for experiment metadata node


singergr:
I believe this was handled in https://github.com/uc-cdis/datadictionary/pull/21.

PR #587, opened 2019-02-26
Fix/setup

jbarno:
travis is mad because attrs isn't getting updated


jbarno:
Yeah, the version file would need to be within the package (i.e. gdcdictionary/VERSION and would need to be valid python) itself so when tox run setup tools it can read it. So I decided setup.py in root was less hidden than `gdcdictionary/__version__.py` Like:
https://github.com/kennethreitz/responder/blob/master/setup.py#L17

I think  https://github.com/warner/python-versioneer would actually be a better long term and automated solution.

gdcdictionary/tests/utils.py
```diff
@@ -15,7 +15,7 @@
 
 def load_yaml(path):
     with open(path, 'r') as f:
-        return yaml.load(f)
+        return yaml.safe_load(f)
```
ProfOak: :clap:

ProfOak: For any future readers:

Even though we probably wouldn't experience malicious uses of `yaml.load()` in this case, given that we aren't running user-submitted yaml, it's still a good practice to use `safe_load()` over `load()`.

https://security.openstack.org/guidelines/dg_avoid-dangerous-input-parsing-libraries.html

PR #638, opened 2019-06-06
DAT-2237: added protein expression node

dmiller15:
https://jira.opensciencedatacloud.org/browse/DAT-2237


dmiller15:
It seems like it should be true. Posted question in the ticket.

PR #691, opened 2019-06-10
feat(qq): bumped version for quilligan quail

dmiller15:


setup.py
```diff
@@ -4,7 +4,7 @@
 
 setup(
     name='gdcdictionary',
-    version="1.16.0",
+    version="1.17.0",
```
kulgan: We already have a `1.17.0`, can you change it to `1.18.0`

PR #745, opened 2019-12-02
chore(release): rc-2.0.0-alpha

kulgan:


setup.py
```diff
@@ -2,7 +2,7 @@
 
 setup(
     name='gdcdictionary',
-    version="1.18.1",
+    version="2.0.0-alpha",
```
singergr: This will work, but the normalized form according to [PEP 440](https://www.python.org/dev/peps/pep-0440/#pre-release-spelling) would be `2.0.0a0` (or we could do `2.0.0a1`).

kulgan: They do allow special separators like `-` or `.`

singergr: Right, it will be accepted like this (using `alpha` in place of `a0` is also allowed), I was just thinking that it would be better to follow the normalized format.

kulgan: This might be harder for the regex used by tools like `bumpversion/versioneer` to auto update. reverting to `2.0.0-a1`

PR #749, opened 2020-02-05
feat/yaml_tsv_tool_update: added few updated files for test purpose

ganjigajanan:
The unit tests are expected to fail, because it needs all the changed files to be part of PR as there are changes which are global. e.g. UUID to uuid


gdcdictionary/schemas/clinical.yaml
```diff
@@ -39,59 +38,48 @@ required:
...
+      by ceil(age*365.25) to get days.'
     type: number
-    description: >
-      Age in days at which a condition or disease was first diagnosed.
-      CDE: 3225640. TCGA has the age in years with CDE 2006657,
-      this is converted by ceil(age*365.25) to get days.
 
   days_to_death:
+    description: "Time interval from a person's date of death to the date of initial\
+      \ pathologic diagnosis, represented as a calculated number of days. CDE: 3165475."
```
philtom-ctds: This is an example where double-quotes are introduced.  It removes internal line breaks with escaping but keeps the line break at the end.

gdcdictionary/schemas/_definitions.yaml
```diff
@@ -44,176 +44,177 @@ to_many:
...
     $ref: "#/md5sum"
+
   file_state:
     $ref: "#/file_state"
+
   error_type:
     $ref: "#/data_file_error_type"
+
   state_comment:
+    description: Optional comment about why the file is in the current state, mainly for invalid state.
```
philtom-ctds: This an example of a single-line, plain string. In the previous version, all line breaks were preserved.  In this version, all line-breaks are removed.

gdcdictionary/schemas/_definitions.yaml
```diff
@@ -1,30 +1,30 @@
 id: _definitions
 
-UUID:
-    term:
-        $ref: "_terms.yaml#/UUID"
-    type: string
-    pattern: "^[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12}$"
+uuid:
+  $ref: "_terms.yaml#/uuid"
```
kulgan: This did not change in the terms file, so should still be `_terms.yaml#/UUID`

gdcdictionary/schemas/aggregated_somatic_mutation.yaml
```diff
@@ -44,39 +45,42 @@ required:
...
+    $ref:
+      - "_terms.yaml#/data_category/common"
+    enum:
+      - Simple Nucleotide Variation
 
   data_type:
-    term:
-      $ref: "_terms.yaml#/data_type"
+    $ref:
+      - "_terms.yaml#/data_type/common"
```
kulgan: Took a closer look and realized that this is going to break the internal resolvers, will need a simple fix to support returning a list

philtom-ctds: Is the cause of failed travis builds?  It's trying to call `split` on a `list`

kulgan: Yea, it was. Looks good now

wwysoc2:
Are the NCIT codes supposed to be in this PR?  Maybe I don't know where to look but I am not seeing them.

ganjigajanan:
@wwysoc2 this is initial yaml migration to new template (standardization). once this once gets merged there will be set of PRs for Thidwick changes.

PR #751, opened 2020-02-14
DAT-2794_DAT-2787:changes to aligned reads

tlicht3:
the ticket number in the branch name (DAT-2705) was incorrect. Changing the ticket description to reflect the correct ticket number (DAT-2794).

### DIFF
--- a/gdcdictionary/schemas/_terms.yaml
+++ b/gdcdictionary/schemas/_terms.yaml
@@ -108,9 +108,10 @@ age_at_index:
 age_is_obfuscated:
   common:
     description: >-
-      The age of the patient has been modified for compliance reasons. The actual
-      age differs from what is reported. Other date intervals for this patient may
-      also be modified.
+      The age or other properties related to the patient's age have been modified
+      for compliance reasons. The actual age may differ from what was reported in
+      order to comply with the Health Insurance Portability and Accountability Act
+      (HIPAA).
     termDef:
       term:
       source:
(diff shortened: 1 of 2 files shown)


gdcdictionary/schemas/aligned_reads.yaml
```diff
@@ -195,6 +196,16 @@ properties:
       - "_terms.yaml#/total_reads/common"
     type: integer
 
+  tumor_ploidy:
+    $ref:
+      - "_terms.yaml#/tumor_ploidy/common"
```
philtom-ctds: I believe the build for the PR is failing because these new properties are referencing the terms file, but the update to the terms files has not happened yet.   There are two options:
1) Add the related changes to `_terms.yaml` to this PR.
2) Submit a separate PR for all of the `_terms.yaml` changes and get it merged first
3) Submit a single PR for all of the changes.

We'll have to choose one approach.  Ideally, we'll do the first option, as that puts all of the related changes together into a single PR.  Then, they can get reviewed together.

However, I'm not sure how the tool works.  If it made all of the yaml changes at once, then the terms files will have a ton of changes, and manually updating each PR to contain only the relevant changes could be a lot of work.  If that is the case, the second option would still allow us submit a single PR per node.  We'll just need to merge all of the terms changes first.   The downside of this one is that it could be difficult to review the terms changes without seeing the corresponding node changes.

The final option is to submit a single PR for all of the changes.  It'll give full context for all of the changes.  The downside is that it'll be a huge PR and will be more difficult to divide up the reviewing responsibilities.

ganjigajanan: This is the first time we are creating PRs using the tool to makes changes. The Thidwick release jira tickets are not structured to help use the TSV<>YAML tool as well as we have taken “Update all at a time approach” in single set of TSVs rather than node specific TSVs to help speedy jira updates & development/updation of tool at the same time.  

We’ll have to choose option 2 to create a PR for "_term.yaml" updates to go in first & then create all the other PRs based off of it, or we can add the fully updated "_terms.yaml" in this PR & get this PR merged & then create other PRs. In the next dictionary release where the dictionary tickets will be per node basis, we can get the node & "_terms.yaml" updated to follow option 1.
 
Also Tara is learning how to create PRs, while we aim at getting changes in sooner & making sure the YAML<> TSV tool does its expected job  (Voila, multitasking at its best! :D), consider the PRs are in the draft mode & please refrain from review until its confirmed to be up for review.

gdcdictionary/schemas/_terms.yaml
```diff
@@ -5327,3 +5328,8212 @@ zygosity:
...
+    termDef:
+      term: >-
+        Six Weeks or More
+      source: NCIt
+      cde_id:
+      cde_version:
+      term_url:
+      term_version: 19.12e
+
+'6':
```
kulgan: This looks a bit off compared to the others. :thinking:

PR #753, opened 2020-02-17
DAT-2746:changes to case node (updated)

tlicht3:
https://jira.opensciencedatacloud.org/browse/DAT-2746


gdcdictionary/schemas/_terms.yaml
```diff
@@ -5584,7 +5584,7 @@ marijuana_use_per_week:
       cde_version:
       term_url:
 
-yes:
+"yes":
```
philtom-ctds: Note to other reviewers:  These two changes to _terms.yaml have already been merged into develop to prevent the other PRs from having issues.

PR #755, opened 2020-02-17
feat/DAT-2808: updates to slide_image node

tlicht3:
https://jira.opensciencedatacloud.org/browse/DAT-2808


tlicht3:
@wwysoc2 yes - I think this must have been an oversight during ticket creation. will make sure this is updated.

PR #756, opened 2020-02-17
DAT-2735: changes to diagnosis node

tlicht3:
https://jira.opensciencedatacloud.org/browse/DAT-2735

### DIFF
--- a/gdcdictionary/schemas/diagnosis.yaml
+++ b/gdcdictionary/schemas/diagnosis.yaml
@@ -83,6 +83,16 @@ properties:
       - Not Allowed To Collect
     deprecated_enum:
       - Not Allowed To Collect
+    enumDef:
+      Unknown:
+        $ref:
+          - "_terms.yaml#/unknown/common"
+      Not Reported:
+        $ref:
+          - "_terms.yaml#/not_reported/common"
+      Not Allowed To Collect:
+        $ref:
+          - "_terms.yaml#/not_allowed_to_collect/common"
 
   ajcc_clinical_n:
     $ref:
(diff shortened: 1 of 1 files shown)


gdcdictionary/schemas/diagnosis.yaml
```diff
@@ -685,8 +942,25 @@ properties:
       - Unknown
       - Not Reported
       - Not Allowed To Collect
+      - Stage IC1
+      - Stage IC2
+      - Stage IC3
+      - Stage IIIA1
+      - Stage IIIAi
+      - Stage IIIAii
+      - Stage IIIA2
```
philtom-ctds: Is there a difference between these?
* `Stage IIIA1` and `Stage IIIAi`
* `Stage IIIA2` and `Stage IIIAii`

I'm asking from a data normalization point of view.  If they are the same, can we pick one over the other?  From the other enums, it looks like arabic numerals are already used for this suffix.

philtom-ctds: I spoke to Bill about this.  They are different things.

gdcdictionary/schemas/diagnosis.yaml
```diff
@@ -504,13 +671,27 @@ properties:
     $ref:
       - "_terms.yaml#/classification_of_tumor/common"
     enum:
-      - primary
       - metastasis
-      - recurrence
       - other
+      - Premalignant
```
wwysoc2: Decided to leave as upper-case as the normalization on the portal will make this all look lower case. We also want to start migrating towards title-case in the dictionary.

PR #775, opened 2020-02-17
feat/DAT-2741:updates to molecular_test node

tlicht3:
https://jira.opensciencedatacloud.org/browse/DAT-2741

### DIFF
--- a/gdcdictionary/schemas/molecular_test.yaml
+++ b/gdcdictionary/schemas/molecular_test.yaml
@@ -85,6 +85,13 @@ properties:
       - Squamous Cell Carcinoma Antigen (SCCA)
       - Unknown
       - Not Reported
+    enumDef:
+      Unknown:
+        $ref:
+          - "_terms.yaml#/unknown/common"
+      Not Reported:
+        $ref:
+          - "_terms.yaml#/not_reported/common"
 
   blood_test_normal_range_lower:
     $ref:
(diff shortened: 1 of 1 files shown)


gdcdictionary/schemas/molecular_test.yaml
```diff
@@ -892,6 +915,17 @@ properties:
       - Unknown
       - Not Reported
       - Not Applicable
+      - MIB1
+      - MAML3 (HCMI)
```
wwysoc2: Is the "HCMI" in DAT-2741 just someone denoting who asked for this gene @tlicht3 ?

gdcdictionary/schemas/molecular_test.yaml
```diff
@@ -980,6 +1028,33 @@ properties:
       - Total Protein
       - Unknown
       - Not Reported
+      - Epstein-Barr Virus
+      - Human Papillomavirus
+      - HPV-E6/E7
+      - Circulating Tumor Cells
+      - B-cell genotyping (BLGSP)
```
wwysoc2: Is the "BLGSP" just signifying which project asked for the change? @tlicht3

gdcdictionary/schemas/molecular_test.yaml
tlicht3: yes - that was an obvious fix - I will make sure this is updated.

gdcdictionary/schemas/molecular_test.yaml
tlicht3: yeah - this also needs updated.

PR #791, opened 2020-03-04
Release/thidwick to develop

philtom-ctds:


gdcdictionary/schemas/somatic_copy_number_workflow.yaml
```diff
@@ -30,7 +30,7 @@ links:
       - name: submitted_genotyping_arrays
         backref: somatic_copy_number_workflows
         label: performed_on
-        target_type: submitted_genotyping_arrays
+        target_type: submitted_genotyping_array
```
ProfOak: `target_type` keys are normally pluralized. Why did we opt for making this singular? It looks like we don't have any of this node in the system so it won't break anything existing, but I'm still curious.

anmaxvl: I think link `name` is pluralized, but not the `target_type`.

philtom-ctds: @ganjigajanan @wwysoc2 Either of you able to answer the question above?

ganjigajanan: @ProfOak, Link `name` & `backref` were always pluralized  & `target_type` refers to the node_type and hence needs to be singular. e.g. `aligned_reads` seems to appear plural but in dictionary context it is the name of the node.

ProfOak: Ah ok my mistake. Thanks for jumping in and answering everyone.

PR #798, opened 2020-06-08
feat/definitions: DICT-26 updates

ganjigajanan:


ganjigajanan:
please review.

gdcdictionary/schemas/aggregated_somatic_mutation.yaml
```diff
@@ -93,7 +93,7 @@ properties:
       - WXS
 
   projects:
-    $ref: "_definitions.yaml#/to_one"
+    $ref: "_definitions.yaml#/to_one_project"
```
kulgan, line 96: Is the plan to progressively move to using `to_one_project`?

gdcdictionary/schemas/_definitions.yaml
```diff
@@ -13,13 +13,43 @@ parent_uuids:
     $ref: "#/uuid"
   uniqueItems: true
 
+foreign_key_project:
```
kulgan, line 16: Just curious if you know how `planX` uses these properties, given `to_one` and `to_many` is generic enough to cover same use cases and these additions only redefine existing properties with a name change.

ganjigajanan: They use "code" instead of "submitter_id" in foreign_key. That's the only difference. Please look up any planX dictionary.

PR #802, opened 2020-06-08
feat/follow_up: DICT-11 updates

ganjigajanan:


ganjigajanan:
please review.

ganjigajanan:
@tlicht3 , uploaded fixed follow_up.yaml file. please review

gdcdictionary/schemas/follow_up.yaml
```diff
@@ -868,12 +868,13 @@ properties:
       - Candidiasis
       - Coccidioidomycosis
       - Cryptococcosis
+      - Cryptosporidiosis, Chronic Intestinal
       - Cytomegalovirus
       - Encephalopathy
       - Herpes Simplex Virus
       - Histoplasmosis
       - Isosporiasis
-      - Mycobacterium avium Complex
```
philtom-ctds, line 876: We may need to deprecate this spelling (casing?).  Changing it is like removing the old value and add the new value.

wwysoc2: Yep, I found one node that uses this old value, let's keep this value here.  I would be fine mirroring this casing across the other fields.  @tlicht3  would this be a problem?

PR #807, opened 2020-06-08
feat/submitted_unaligned_reads: DICT-13 updates

ganjigajanan:


ganjigajanan:
please review.

gdcdictionary/schemas/submitted_unaligned_reads.yaml
```diff
@@ -90,7 +90,6 @@ properties:
     $ref:
       - "_terms.yaml#/read_pair_number/common"
     enum:
-      - I1
```
kulgan, line 93: If we ever had this value, it would have broken export to bio. :+1:

PR #810, opened 2020-06-08
Feat/diagnosis: DICT-6 updates

tlicht3:
These updates are described here: https://jira.opensciencedatacloud.org/browse/DICT-6


gdcdictionary/schemas/diagnosis.yaml
```diff
@@ -35,22 +35,21 @@ links:
 required:
   - submitter_id
   - age_at_diagnosis
-  - days_to_last_follow_up
```
kulgan, line 38: need to update the `diagnosis.json` sample in the examples folder

kulgan: :facepalm: these are just no longer required

gdcdictionary/schemas/diagnosis.yaml
```diff
@@ -804,6 +1610,24 @@ properties:
         minimum: -32872
       - type: "null"
 
+  eln_risk_classification:
```
kulgan, line 1613: Shouldn't this have a `$ref` entry?

gdcdictionary/schemas/diagnosis.yaml
```diff
@@ -6007,13 +17325,773 @@ properties:
...
       Unknown:
         $ref:
           - "_terms.yaml#/unknown/common"
       Not Reported:
         $ref:
           - "_terms.yaml#/not_reported/common"
 
+  sites_of_involvement:
+    type: array
+    items: enum
```
kulgan, line 18093: need to give enum values

PR #811, opened 2020-06-09
feat/sample: DICT-7 updates

tlicht3:
https://jira.opensciencedatacloud.org/browse/DICT-7


wwysoc2:
It looks like this PR for DICT-7 has changes for DICT-6 and DICT-16  too

PR #815, opened 2020-06-26
reverted DICT-5 - RMA downloadable

wwysoc2:


gdcdictionary/schemas/raw_methylation_array.yaml
```diff
@@ -11,7 +11,7 @@ description: >-
   Data file containing channel data from a methylation array.
 additionalProperties: false
 submittable: true
-downloadable: false
+downloadable: true
```
philtom-ctds, line 14: There's a trailing space after `true`.  I'm not sure if it'll affect anything downstream, but it'll be safer to remove it.

wwysoc2: Good catch!  Should be gone now.

PR #822, opened 2020-11-23
DICT-64: Updates to the TERMS file - YERTLE

tlicht3:
This PR represents the changes documented in this ticket: https://jira.opensciencedatacloud.org/browse/DICT-64

In addition to the change to **aliquot_volume**, new terms have been added for enumerations on the molecular_test and the pathology_detail nodes. These new terms have been added at the end of the file.


wwysoc2:
I was able to verify that all of the enums from pathology_detail and molecular_test are present in the _terms.yaml file.  I'm not sure how much further we want to take testing, since this is a massive file.

ganjigajanan (approved):

PR #825, opened 2020-11-30
DICT-57: Updates to the TREATMENT node - YERTLE

tlicht3:
For a complete list of changes refer to this ticket: https://jira.opensciencedatacloud.org/browse/DICT-57

### DIFF
--- a/gdcdictionary/schemas/treatment.yaml
+++ b/gdcdictionary/schemas/treatment.yaml
@@ -431,9 +431,36 @@ properties:
       - Unknown
       - Not Reported
     enumDef:
+      Bone Marrow:
+        $ref:
+          - "_terms.yaml#/bone_marrow/common"
+      Buccal Mucosa:
+        $ref:
+          - "_terms.yaml#/buccal_mucosa/common"
+      Cerebrospinal Fluid:
+        $ref:
+          - "_terms.yaml#/cerebrospinal_fluid/common"
+      Connective Tissue:
+        $ref:
+          - "_terms.yaml#/connective_tissue/common"
+      Epitrochlear:
+        $ref:
+          - "_terms.yaml#/epitrochlear/common"
+      Frontal Lobe:
+        $ref:
+          - "_terms.yaml#/frontal_lobe/common"
+      Hilar:
+        $ref:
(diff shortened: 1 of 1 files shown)


gdcdictionary/schemas/treatment.yaml
```diff
@@ -634,9 +666,12 @@ properties:
       - Radiation, Combination
       - Radiation, Cyberknife
       - Radiation, External Beam
+      - Radiation, Hypofractionated
       - Radiation, Implants
       - Radiation, Intensity-Modulated Radiotherapy
       - Radiation, Internal
+      - Radiation, Mixed Photon Beam
+      - Radiation, Photon Beam
```
wteouchicago, line 674: not sure if it matters, but _Radiation, Photon Beam_ shows up twice (lines 674 and 675)

gdcdictionary/schemas/treatment.yaml
```diff
@@ -3417,10 +3854,11 @@ properties:
...
       - Oregovomab
+      - Orelabrutinib
       - Ormaplatin
       - Ortataxel
       - Orteronel
-      - OSI-774 (erlotinib; Tarceva)
+      - Orvacabtagene Autoleucel
       - Osilodrostat
       - Osimertinib
       - Other
```
wteouchicago, line 3864: Do we need to remove "Other" here (in addition to removing it from the deprecated_enum section)?

tlicht3: @wteouchicago we are not removing "Other". Instead, we are removing it from the deprecated list, as we need it to stay on the list of values.

gdcdictionary/schemas/treatment.yaml
tlicht3: @wteouchicago Proton and Photon are two different types of radiation, so we need both of them. Thanks for checking on this!

wteouchicago: Ah thanks! I clearly need my eyes checked!

wteouchicago (approved):

wwysoc2 (approved):

PR #838, opened 2020-12-03
DICT-54: Updates to the DIAGNOSIS node - YERTLE

tlicht3:
For complete details, see this ticket: https://jira.opensciencedatacloud.org/browse/DICT-54

### DIFF
--- a/gdcdictionary/schemas/diagnosis.yaml
+++ b/gdcdictionary/schemas/diagnosis.yaml
@@ -241,6 +241,7 @@ properties:
       - Stage IA
       - Stage IA1
       - Stage IA2
+      - Stage IA3
       - Stage IB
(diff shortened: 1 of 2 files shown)


gdcdictionary/schemas/diagnosis.yaml
```diff
@@ -18106,7 +18208,11 @@ properties:
     enum:
```
wwysoc2, line 18208: I'm not sure if I'm missing something here, but  I don't see the "supratentorial_localization" property
--

wwysoc2 (changes requested):
Everything looks good except that supratentorial_localization seems to be missing.

wwysoc2 (approved):

PR #846, opened 2021-01-12
Feat/dev 497 prevent duplicate enum values

XukunXiangUChicago:
go through all property `enum` and report all the duplicates

### DIFF
--- a/tests/schema_test.py
+++ b/tests/schema_test.py
@@ -13,6 +13,15 @@ from .utils import BaseTest, check_for_cycles, validate_schemata
 
 
 def _check_enum(enums):
+    """Check the collected enums
+
+    Args:
+        enums (dict of str: list[str]): mapping from the path to the enum to the enum object
+
+    Returns:
+        dict: mapping from the violations and the corresponding path (and item name if necessary)
+
+    """
     check_result = collections.defaultdict(list)
     for path, enum_list in enums.items():
         if not isinstance(enum_list, list):
@@ -31,6 +40,15 @@ def _check_enum(enums):
 
 
 def _generate_error_message_for_enum(res_dict):
+    """Generate error message string for enum checks
+
+    Args:
+        res_dict (dict of string: string): mapping from the violation to the list of enum path
+
+    Returns:
+         str: the formatted error message
+
+    """
     message_lines = ["Errors in enum checks:"]
     for error_name, lines in res_dict.items():
         message_lines.append(error_name)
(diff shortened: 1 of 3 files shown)


jiakf (approved):
lgtm!

tests/schema_test.py
```diff
@@ -6,37 +6,60 @@
...
-        elif isinstance(root, dict):
-            for child in root.keys():
-                self.traverse_schema_with_conditional(root[child], conditional_func, test_func)
+        def _get_path_str(path_stack):
+            return '->'.join(path_stack)
 
-    def test_enum_is_string(self):
-
-        def trigger(root):
+        def bfs_with_enum_test(root, test_func, path):
```
BenX00, line 25: Couldn't we make this a static class method? To me these functions encapsulated in the test function add a lot of clutter and make it hard to read. Also would you be able to give them a doc string? bfs_with_enums just reads as best friends with enums and not exactly sure what this is doing lol :P 

Also this approach relies completely on side effect with all of these methods building on and appending common objects. I'd challenge you to take a more functional approach build things up and have meaningful returns with no side effect outside of their own scope.

kulgan: I don't think bfs is the right terminology for going through the node definitions in `self.dictionary.schema`. It doesn't have to be recursive too, I am guessing someone just wanted to be fancy when they did the old implementation. Some other way to look at this would be something like this
``` python
for node, node_dict in self.dictionary.schema.items():
    for prop, prop_dict in node_dict["properties"].items():
        if "enum" in prop_dict:
            # do some magic
```

kulgan (changes requested):

tests/schema_test.py
```diff
@@ -6,37 +6,53 @@
...
-            for child in children:
-                assert isinstance(child, str), "Offending enum found:" + str(child)
-
-        self.traverse_schema_with_conditional(self.dictionary.schema, trigger, test_enum)
+    @unittest.expectedFailure
+    def test_properties_enum(self):
+        """Check the enums of node properties"""
+        # The enums in _definitions.yaml, _terms.yaml and metaschema.yaml are not checked
+
+        def check_enum(enums):
```
BenX00, line 22: This reads a lot better. I still thing that this function and the one below are better suited as methods in the test class itself.

XukunXiangUChicago: It seems that we prefer module level functions over static methods and class methods, so I moved them outside of the class (ref: https://google.github.io/styleguide/pyguide.html#217-function-and-method-decorators)

kulgan (approved):

BenX00 (approved):
:+1:

PR #847, opened 2021-01-13
DEV-489 add test for checking links with annotation

XukunXiangUChicago:
the test checks the `links.subgroup` section of `annotation.yaml` to verify that appropriate nodes are linked with annotation nodes

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -66,6 +66,7 @@ target/
 
 #venv
 .venv/
+venv*/
 
 #DS_store
 .DS_store
@@ -88,5 +89,6 @@ admin_auth
 # vim
 *.swp
 
-# vscode
+# IDEs
 .vscode/
+.idea/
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -88,5 +89,6 @@ admin_auth
 # vim
 *.swp
 
-# vscode
+# IDEs
```
ProfOak, line 92: :100:

tests/schema_test.py
```diff
@@ -154,3 +154,29 @@ def test_check_for_cycles_negative(self):
 
         with self.assertRaisesRegexp(AssertionError, 'cycle detected'):
             check_for_cycles(schemata)
+
+    def test_links_to_annotation(self):
+        special_node = [
+            'annotation', 'clinical', 'experimental_strategy', 'metaschema',
+            'platform', 'program', 'project', 'publication', 'root', 'tag'
+        ]
```
ProfOak, line 162: It's a minor thing but if we use a set here, the `x in set` that we do in `_is_excluded_node` will be constant time instead of linear.

ProfOak (approved):
Seems good to me. I left a comment but I don't think it warrants blocking the merge.

kulgan (approved):

scintillating7 (approved):

PR #857, opened 2021-06-02
boveri Exposure node updates

ganjigajanan:


gdcdictionary/schemas/exposure.yaml
```diff
@@ -229,11 +235,17 @@ properties:
         $ref:
           - "_terms.yaml#/not_reported/common"
 
+  exposure_duration_years:
+    $ref:
+      - "_terms.yaml#/exposure_duration_years/common"
+    type: integer
+
```
wteouchicago, line 242: we might be able to add a min of 0 here.

wteouchicago:
looks good according to DICT-94

gdcdictionary/schemas/exposure.yaml
wwysoc2: +1

wwysoc2 (changes requested):

gdcdictionary/schemas/exposure.yaml
ganjigajanan: added

wwysoc2 (approved):

PR #859, opened 2021-06-02
boveri Pathology_detail node updates

ganjigajanan:


gdcdictionary/schemas/pathology_detail.yaml
```diff
@@ -46,7 +46,6 @@ properties:
   additional_pathology_findings:
     $ref:
       - "_terms.yaml#/additional_pathology_findings/common"
```
wwysoc2, line 48: We need the enums to be here @ganjigajanan

gdcdictionary/schemas/pathology_detail.yaml
```diff
@@ -46,7 +46,6 @@ properties:
   additional_pathology_findings:
```
wteouchicago, line 46: DICT-96 says there are enums added:
Pleurodesis, Talc
Pleurodesis, NOS

I don't see them in the changes made.

wwysoc2 (changes requested):

gdcdictionary/schemas/pathology_detail.yaml
ganjigajanan: added the enums

gdcdictionary/schemas/pathology_detail.yaml
ganjigajanan: added

wwysoc2 (approved):

PR #865, opened 2021-06-02
boveri Somatic_aggregation_workflow node updates

ganjigajanan:


mforootan:
Looks OK

gdcdictionary/schemas/somatic_aggregation_workflow.yaml
```diff
@@ -24,12 +24,21 @@ systemProperties:
...
-    multiplicity: one_to_many
+  - exclusive: false
     required: true
+    subgroup:
+      - name: annotated_somatic_mutations
+        backref: somatic_aggregation_workflows
+        label: performed_on
+        target_type: annotated_somatic_mutation
+        multiplicity: one_to_many
+        required: true
```
wwysoc2, line 35: This should be "false" I think, since we want either to work.

gdcdictionary/schemas/somatic_aggregation_workflow.yaml
```diff
@@ -24,12 +24,21 @@ systemProperties:
   - state
 
 links:
-  - name: annotated_somatic_mutations
-    backref: somatic_aggregation_workflows
-    label: performed_on
-    target_type: annotated_somatic_mutation
-    multiplicity: one_to_many
+  - exclusive: false
```
wwysoc2, line 27: Change to True

wwysoc2 (changes requested):

gdcdictionary/schemas/somatic_aggregation_workflow.yaml
ganjigajanan: updated

gdcdictionary/schemas/somatic_aggregation_workflow.yaml
ganjigajanan: updated

wwysoc2 (approved):

PR #873, opened 2021-06-02
boveri Analyte node updates

ganjigajanan:


gdcdictionary/schemas/analyte.yaml
```diff
@@ -111,11 +114,13 @@ properties:
     $ref:
       - "_terms.yaml#/analyte_volume/common"
     type: number
+    minimum: 0
```
wteouchicago, line 117: didn't spot this in the ticket (DICT-100), but it makes sense.

gdcdictionary/schemas/analyte.yaml
```diff
@@ -168,11 +173,13 @@ properties:
     $ref:
       - "_terms.yaml#/ribosomal_rna_28s_16s_ratio/common"
     type: number
+    minimum: 0
 
   rna_integrity_number:
```
wteouchicago, line 178: ticket has a min (DICT-100)

gdcdictionary/schemas/analyte.yaml
ganjigajanan: confirmed with Bill as well

wteouchicago (changes requested):
well_number changes are in the ticket but not here.

gdcdictionary/schemas/analyte.yaml
wteouchicago: oops - that was for rna_integrity_number

ganjigajanan: Confirmed with Bill the rna_integrity_number minimum value mentioned in ticket needs review. hence did not add it to node. can be added in future if needed after confirming

wwysoc2 (approved):

wwysoc2 (approved):

wwysoc2:
We decided not to do the well number changes.  @wteouchicago could you consider the changes done?  I can't figure out how to do that.

wteouchicago (approved):
my concerns have been addressed.

PR #876, opened 2021-06-03
boveri _terms file split into 2

ganjigajanan:


philtom-ctds:
at a minimum, you'll need to add the new terms file here to prevent other code from interpreting it as a node:
https://github.com/NCI-GDC/gdcdictionary/blob/develop/gdcdictionary/python/__init__.py#L34-L37

wteouchicago (approved):
glanced at _terms.yaml and _terms_enum.yaml and looked over most of the <nodes>.yaml files. didn't spot anything out of place.

wwysoc2 (approved):
lgtm

tlicht3:
@gajanan - why are there only 101,404 additions to the terms_enum file, but 102,446 deletions on the terms file? Did you remove some that aren't being referenced or some other type of clean-up? Really just curious.

PR #877, opened 2021-06-03
boveri _terms update the desc of percent_lymphocyte_infiltration

ganjigajanan:


wwysoc2 (approved):
lgtm

mforootan:
Looks OK as we planned, just for the record, I paste Nicollette M's comments as well:
A majority of literature references lymphocyte infiltration readings RE tumor samples.
Many of the GDC data dictionary infiltration properties also reference tumor sample in their definitions.
Perhaps specifying "tumor" samples is more appropriate for this property as well?

tlicht3:
Thanks, Massih... we could probably update all of these to remove "tumor"
as these different cell types could show up in normal and tumor samples.
Something to think about for future releases, but nothing urgent that would
require a last-minute update.

PR #878, opened 2021-06-04
fix(tests): jenkins build with minor updates

kulgan:
Jenkins fails with unicode encoding errors, claiming some non ascii characters where found while loading the dictionary. The schemas are loaded using `open` without specifying an encoding, making it platform dependent. The jenkins errors are probably due to jenkins tests using a different default encoding. This PR sets the encoding to utf-8 while loading the dictionary. It uses the `codecs` module since python2.7 open does not allow for specifying encoding

### DIFF
--- a/gdcdictionary/python/__init__.py
+++ b/gdcdictionary/python/__init__.py
@@ -1,3 +1,4 @@
+import codecs
 from copy import deepcopy
 from collections import namedtuple
 from contextlib import contextmanager
@@ -65,7 +66,7 @@ class GDCDictionary(object):
         """Return contents of yaml file as dict"""
         # For DAT-1064 Bomb out hard if unicode is in a schema file
         # But allow unicode through the terms and definitions
-        with open(name, 'r', encoding="utf-8") as f:
+        with codecs.open(name, 'r', encoding="utf-8") as f:
             if name not in self.exclude:
                 try:
                     f.read().encode("ascii")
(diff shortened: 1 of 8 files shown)


gdcdictionary/python/__init__.py
```diff
@@ -65,7 +66,7 @@ def load_yaml(self, name):
         """Return contents of yaml file as dict"""
         # For DAT-1064 Bomb out hard if unicode is in a schema file
         # But allow unicode through the terms and definitions
-        with open(name, 'r') as f:
+        with codecs.open(name, 'r', encoding="utf-8") as f:
```
philtom-ctds, line 69: The problem is in _terms.yaml.
```
+route_of_administration:
+      <U+200B>The pathway by which a substance is administered in order to reach the site
+      of action in the body.
+        <U+200B>Route of Administration
+      cde_id: C38114<U+200B>
+      cde_version: 21.04d
+      term_url: "https://ncit.nci.nih.gov/ncitbrowser/ConceptReport.jsp?dictionary=NCI_Thesaurus&ns=ncit&code=C38114"
```
There are now zero-width whitespace characters (U+200B).  They should be deleted.

ProfOak: Nice catch. I wonder if a word processor inserted these characters.

kulgan: I tried to dig into the source behind the comment about allowing unicode in terms and definitions but found nothing

kulgan:
No longer needed

PR #879, opened 2021-06-04
boveri _terms update - remove zero width whitespace chars

ganjigajanan:


ProfOak (approved):
I wonder if this could have been avoided if we enabled the [pre-commit](https://pre-commit.com/#2-add-a-pre-commit-configuration) `trailing-whitespace`.

wwysoc2 (approved):

PR #900, opened 2022-06-14
SomaticAnnotationWorkflow changes - Darwin Release

wwysoc2:


wwysoc2 (commented):
enum changes lgtm, the nested $term and $ref need to be explored further:  @tlicht3  @ganjigajanan

PR #910, opened 2022-06-15
_terms changes - Darwin Release

wwysoc2:


mpsolano (commented):
This didn't include the changes to test_value or analyte_type_id referenced in DICT-179. Not sure if those were addressed in another ticket or if they should be included here.

wteouchicago (commented):
Noted the same things Mike found:
- test_value update missing
- HARRT => HAART in termDef
- it looks like from DICT-179 that we don't need to add the description for analyte_type_id anymore

wwysoc2:
@tlicht3 had a bunch of changes which makes the term file way different from what I put in this PR.  I'm going to close this one and make a new PR.

PR #916, opened 2022-06-16
Exposure changes - Darwin Release

wwysoc2:


gdcdictionary/schemas/exposure.yaml
```diff
@@ -38,15 +38,29 @@ uniqueKeys:
...
+    $ref:
+      - "_terms.yaml#/asbestos_exposure_type/common"
+    enum:
+      - Amosite
+      - Crocidolite
+
+  age_at_last_exposure:
+    $ref:
+      - "_terms.yaml#/age_at_last_exposure/common"
+    type: integer
```
wwysoc2, line 62: Need to add min/max here.

wwysoc2:
Changes look good to me.

gdcdictionary/schemas/exposure.yaml
```diff
@@ -38,15 +38,31 @@ uniqueKeys:
...
 
 deprecated:
-  - bmi
-  - height
-  - weight
+  - smokeless_tobacco_quit_age
+  - asbestos_exposure
+  - radon_exposure
+  - respirable_crystalline_silica_exposure
+  - years_smoked
```
wteouchicago, line 45: are we supposed to be removing years_smoked or deprecating it?

gdcdictionary/schemas/exposure.yaml
```diff
@@ -704,6 +716,11 @@ properties:
         $ref:
           - "_terms_enum.yaml#/smokeless_tobacco/common"
 
+  use_per_day:
```
wteouchicago, line 719: not in the ticket, but should/can we do a min: 0 for this one?

gdcdictionary/schemas/exposure.yaml
tlicht3: @wwysoc2 said that the migration for this property wasn't done, so instead of removing it we need to add it to the deprecated list.

mpsolano (approved):
Looks good!

PR #919, opened 2022-07-19
feat: use the c library to safeload yaml files

ProfOak:
I tested this by loading the `_terms_enum.yaml` file, which is the largest yaml file in this repo. `safe_load` took just over 7 seconds to complete. Using `CSafeLoader` as the loader took just over 1 second to complete.

`CSafeLoader` can only be imported if libyaml is installed locally. If it still says that `CSafeLoader` cannot be imported then delete the pip wheel cache. The feature is only enabled when pip installs the library, not when the module is imported, which is sort of confusing.

Linux:
```~/.cache/pip```

Mac:
```~/Library/Caches/pip```

Windows:
```%LocalAppData%\pip\Cache```

### DIFF
--- a/gdcdictionary/python/__init__.py
+++ b/gdcdictionary/python/__init__.py
@@ -73,7 +73,7 @@ class GDCDictionary(object):
                 except Exception as e:
                     self.logger.error("Error in file: {}".format(name))
                     raise e
-            return yaml.safe_load(f)
+            return yaml.load(f, Loader=yaml.CSafeLoader)
 
     def load_schemas_from_dir(self, directory):
         """Returns all yamls and resolvers of those yamls from dir"""
--- a/gdcdictionary/python/__init__.py
+++ b/gdcdictionary/python/__init__.py
@@ -73,7 +73,12 @@ class GDCDictionary(object):
                 except Exception as e:
                     self.logger.error("Error in file: {}".format(name))
                     raise e
-            return yaml.load(f, Loader=yaml.CSafeLoader)
+            try:
+                return yaml.load(f, Loader=yaml.CSafeLoader)
+            except AttributeError as e:
+                self.logger.warning(e)
+                self.logger.warning("To enable CSafeLoader install libyaml. Falling back to yaml.safe_load()")
+                return yaml.safe_load(f)
 
     def load_schemas_from_dir(self, directory):
         """Returns all yamls and resolvers of those yamls from dir"""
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: gdcdictionary/python/__init__.py [hunk 1]
```diff
@@ -73,7 +73,7 @@ def load_yaml(self, name):
                 except Exception as e:
                     self.logger.error("Error in file: {}".format(name))
                     raise e
-            return yaml.safe_load(f)
+            return yaml.load(f, Loader=yaml.CSafeLoader)
```
kulgan, line 76: 👍🏾

kulgan (approved):

jiakf (approved):

XukunXiangUChicago (approved):

gdcdictionary/python/__init__.py
```diff
@@ -73,7 +73,12 @@ def load_yaml(self, name):
                 except Exception as e:
                     self.logger.error("Error in file: {}".format(name))
                     raise e
-            return yaml.safe_load(f)
+            try:
+                return yaml.load(f, Loader=yaml.CSafeLoader)
+            except AttributeError as e:
+                self.logger.warning(e)
+                self.logger.warning("To enable CSafeLoader install libyaml. Falling back to yaml.safe_load()")
+                return yaml.safe_load(f)
```
philtom-ctds, line 81: pyyaml sets a [flag](https://github.com/yaml/pyyaml/blob/0abad85a17ba75c0fb431feea7a6a06125341a99/lib/yaml/__init__.py#L12-L16
) to indicate if libyaml extension is available.  let's use it:
```suggestion
            if yaml.__with_libyaml__:
                return yaml.load(f, Loader=yaml.CSafeLoader)
            else:
                return yaml.safe_load(f)
```

gdcdictionary/python/__init__.py
```diff
@@ -73,6 +73,10 @@ def load_yaml(self, name):
                 except Exception as e:
                     self.logger.error("Error in file: {}".format(name))
                     raise e
+            if yaml.__with_libyaml__:
+                return yaml.load(f, Loader=yaml.CSafeLoader)
+            self.logger.warning(e)
```
philtom-ctds, line 78: there's no longer an exception.  remove this line.

gdcdictionary/python/__init__.py
```diff
@@ -73,6 +73,10 @@ def load_yaml(self, name):
...
+            self.logger.warning("To enable CSafeLoader install libyaml. Falling back to yaml.safe_load()")
```
philtom-ctds, line 79: A warning means that something went wrong.  Nothing is wrong here.  `debug` would be an appropriate log level.

philtom-ctds (approved):

PR #928, opened 2023-01-17
update follow_up node DICT-224

bullenc:


gdcdictionary/schemas/follow_up.yaml
```diff
@@ -956,9 +956,9 @@ properties:
         $ref:
           - "_terms_enum.yaml#/not_reported/common"
 
-  comorbidity:
+  comorbidities:
```
wteouchicago, line 959: question: are we formally deprecating 'comorbidity' or just kind of deleting it? if we are trying to formally deprecate, do we need to add it to the `deprecated` list instead?

wteouchicago:
I am guessing this PR is still in progress, but just in case: quite a few things in the jira are not in, e.g. adding new properties timepoint_category, imaging_findings, etc.

gdcdictionary/schemas/follow_up.yaml
bullenc: added to deprecated properties, kept original entry.

gdcdictionary/examples/valid/follow_up.json
```diff
@@ -5,7 +5,7 @@
     "barretts_esophagus_goblet_cells_present":"No",
     "bmi":10,
     "cause_of_response":"First Line of Therapy",
-    "comorbidity":"COPD",
+    "comorbidities":"COPD",
```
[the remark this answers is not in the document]
bullenc: ahhh forgot about this, thanks

gdcdictionary/schemas/follow_up.yaml
```diff
@@ -1580,6 +1804,24 @@ properties:
...
+      enum:
+        - Abdomen
+        - Pelvis
+
+  imaging_findings:
+    enum:
+      - Carcinomatosis
+      - Kidney Involvement
+      - Lung Involvement
+      - Liver Involvement
```
wteouchicago, line 1819: liver before lung. :P

wteouchicago, line 1819: liver should be above lung. :P

gdcdictionary/schemas/follow_up.yaml
```diff
@@ -1580,6 +1804,24 @@ properties:
...
+      - Normal
+      - Vena Cava Involvement/Thrombus
+      - Retroperitoneal Lymph Node Involvement
```
wteouchicago, line 1822: Retro...above Vena....

gdcdictionary/schemas/follow_up.yaml
```diff
@@ -956,9 +958,196 @@ properties:
...
+        - Urinary Tract Infection
+        - Varicella Zoster Virus
+        - Wagr Syndrome
+        - Unknown
+        - Not Reported
+
   comorbidity:
     $ref:
-      - "_terms.yaml#/comorbidity/common"
+      - "_terms.yaml#/comorbidities/common"
```
wteouchicago, line 1150: qn. if we are deprecated comorbidity, is it better to remove the ref to terms or update the ref (which you do here)?

gdcdictionary/schemas/follow_up.yaml
```diff
@@ -1185,6 +1377,9 @@ properties:
     maximum: 32872
     minimum: -32872
 
+  days_to_first_event:
+    type: integer
+
```
wteouchicago, line 1382: feels like we can add a max and min

gdcdictionary/schemas/follow_up.yaml
```diff
@@ -2125,6 +2383,7 @@ properties:
       - Regional
       - Unknown
       - Not Reported
+      - Locoregional
```
wteouchicago, line 2386: alphabetize

gdcdictionary/schemas/follow_up.yaml
```diff
@@ -1693,6 +1941,15 @@ properties:
     type: integer
     minimum: 1900
 
+  peritoneal_washing_results:
+    enum:
+      - Positive
+      - Negative
```
wteouchicago, line 1947: do we need to alphabetize here?

gdcdictionary/schemas/follow_up.yaml
```diff
@@ -1693,6 +1941,15 @@ properties:
...
+      - Not Reported
+
+  pregnancy_count:
+    type: integer
```
wteouchicago, line 1951: would a min of 0 be appropriate?

gdcdictionary/schemas/follow_up.yaml
```diff
@@ -1704,6 +1961,7 @@ properties:
       - Stillbirth
       - Unknown
       - Not Reported
+      - Spontaneous Abortion
```
wteouchicago, line 1964: alphabetize

gdcdictionary/schemas/follow_up.yaml
```diff
@@ -1604,6 +1846,12 @@ properties:
         $ref:
           - "_terms_enum.yaml#/not_reported/common"
 
+  imaging_suv:
+    type: number
+
+  imaging_suv_max:
+    type: number
+
```
wteouchicago, line 1854: qn. are there appropriate mins or maxes here? like maybe a min of 0?

gdcdictionary/schemas/follow_up.yaml
```diff
@@ -1185,6 +1377,9 @@ properties:
     maximum: 32872
     minimum: -32872
 
+  days_to_first_event:
```
mpsolano, line 1380: Do this and the other new properties need a `$ref`?

gdcdictionary/schemas/follow_up.yaml
```diff
@@ -2170,8 +2429,6 @@ properties:
           - "_terms_enum.yaml#/not_reported/common"
 
   risk_factor:
-    $ref:
```
mpsolano, line 2173: Here we are removing the terms ref but for comorbidity we are changing it from comorbidity to comorbidities. Is there a reason we are treating risk_factor differently? I would expect it to change similarly (i.e. risk_factor -> risk_factors).

gdcdictionary/schemas/follow_up.yaml
wwysoc2: Yes, this should be resolved by now.

gdcdictionary/schemas/follow_up.yaml
```diff
@@ -1321,15 +1525,22 @@ properties:
         $ref:
           - "_terms_enum.yaml#/not_reported/common"
 
+  evidence_of_progression_type:
```
wwysoc2, line 1530: @tlicht3 , should this be here?  I don't see it in the JIRA ticket?  Also if it is needed, it should have an entry in the terms file.

gdcdictionary/schemas/follow_up.yaml
wwysoc2: Fixed.

gdcdictionary/schemas/follow_up.yaml
wwysoc2: Fixed

gdcdictionary/schemas/follow_up.yaml
wwysoc2: Fixed

gdcdictionary/schemas/follow_up.yaml
wwysoc2: Yes!  Fixed.

gdcdictionary/schemas/follow_up.yaml
wwysoc2: Good point, not technically necessary (since risk_factor is being deprecated), but added ref for consistency.

gdcdictionary/schemas/follow_up.yaml
bullenc: I updated the ref to get the travis checks to pass since I think comorbidity was removed from terms file. Not sure what the precedent is for this, I can try removing the ref and see if it makes a difference at all.

gdcdictionary/schemas/follow_up.yaml
bullenc: Fixed

gdcdictionary/schemas/follow_up.yaml
wwysoc2: Fixed.

gdcdictionary/schemas/follow_up.yaml
wwysoc2: Fixed.

gdcdictionary/schemas/follow_up.yaml
wwysoc2: I read the description and I have no idea.  @tlicht3 do you know offhand?  I don't suggest we put too much research into this.

mpsolano (approved):

gdcdictionary/schemas/follow_up.yaml
```diff
@@ -1321,15 +1525,24 @@ properties:
...
+    $ref:
+      - "_terms.yaml#/evidence_of_progression_type/common"
```
mpsolano, line 1530: I don't think there is a corresponding entry in `_terms`.yaml for `evidence_of_progression_type` which may be the reason the build is failing

mpsolano (approved):
This looks good, just had one suggestion for fixing the build issue

gdcdictionary/schemas/follow_up.yaml
tlicht3: @bullenc and @wteouchicago I'm a little confused... we're looking at the old property here, right - the deprecated property? I think in the past we have left that as is (outside of adding it to the deprecated list) until it's ready to be removed (i.e. after the data migration to the new property). Not sure if that's helpful, but I do think we should create a new entry in the terms file for the plural version so that it matches the follow-up schema.

wteouchicago: @tlicht3 I'm not 100% familiar with what we did the past, but yeah, I had expected something like:
- just add the old singular comorbidity to deprecated list
- add the new plural comorbidities with all its enums and also to terms file

But I defer to everyone else on this. :)

gdcdictionary/schemas/follow_up.yaml
tlicht3: @wwysoc2 and @wteouchicago I think we can ask EVS for this when we add the term to NCIt... after reading about it, I agree that it's unlikely to be negative, but I don't know enough about it to know for sure.

gdcdictionary/schemas/follow_up.yaml
tlicht3: @wwysoc2 ah - this was a discussion point in that long email from Megan. I think we need this because this is how the recurrence type is collected. I'll look into this and will update the terms file with this and the kidney staging one that needs added.

gdcdictionary/schemas/follow_up.yaml
tlicht3: @wwysoc2 and @mpsolano I'll update this.

gdcdictionary/schemas/follow_up.yaml
bullenc: @tlicht3 and @wteouchicago so comorbidity is in the deprecated list, context for the above changes: I just changed (then deleted) the reference for the deprecated comorbidity property since the _comorbidity_ entry in the _terms.yaml file was removed- I assumed this was intentional; if so, checks would break because the property in the follow_up.yaml file pointed to a non-existent term in the _terms.yaml file- so thus I changed to an existing one ("_terms.yaml#/comorbidities/common") for checks to go thru, and then I just removed it altogether after Wendy's comment. 

It sounds like the removal of the comorbidity entry in the _terms.yaml file was erroneous and precedent was to leave everything alone for deprecated properties.

Perhaps then we should:

> add back in entry for  comorbidity in _terms.yaml file
> add back in "_terms.yaml#/comorbidity/common" reference for comorbidity

wteouchicago: that sounds right to me, though @wwysoc2 can probably confirm as well.

gdcdictionary/schemas/follow_up.yaml
mpsolano: 👍

gdcdictionary/schemas/follow_up.yaml
wwysoc2: Do any of these options that you would like (or leave it as is).  As long as "comorbidities" points to "comorbidities", that's okay with me.  "comorbidity" is deprecated, and therefore won't show up in the viewer, so it shouldn't matter where the ref file points.

tlicht3: this all makes sense to me - i thought we needed the old term for the test to pass... i think i now understand that if we have the ref and there's nothing to reference in terms file it will error out, but there's nothing wrong with removing BOTH the ref and the term in the terms file.

bullenc: Ok I will leave it as is. Thanks for the input all!

gdcdictionary/schemas/follow_up.yaml
```diff
@@ -2373,6 +2867,28 @@ properties:
       - PSMA
       - Sodium Fluoride
 
+  timepoint_category:
+    $ref:
+      - "_terms.yaml#/timepoint_category/common"
+    enum:
+      - Adolescence
+      - Adjuvant therapy
```
wwysoc2, line 2875: Seems to be sentence case, rather than title case.

wwysoc2: I'll change the JIRA ticket to reflect this.

gdcdictionary/schemas/follow_up.yaml
```diff
@@ -1185,6 +1377,13 @@ properties:
     maximum: 32872
     minimum: -32872
 
+  days_to_first_event:
+    $ref:
+      - "_terms.yaml#days_to_first_event"
```
wwysoc2, line 1382: Does this have a "common" after it?  (I have no idea if this is required or something)

gdcdictionary/schemas/follow_up.yaml
bullenc: Good catch, updated

gdcdictionary/schemas/follow_up.yaml
bullenc: Yes I think so, updated it

wwysoc2 (approved):
👍

PR #944, opened 2023-01-17
Remove extraneous unicode characters from _terms file

ProfOak:


qiaouchicago (approved):

philtom-ctds (approved):
Looks good.  I double checked the rest of the files.  They are all ASCII:
```
$ file gdcdictionary/schemas/*yaml | grep -v ASCII | wc -l
0
```
(^ that command say "none are not ascii", which is the same as "all are ascii")

PR #961, opened 2023-02-06
DEV-1672 automate gdcdatamodel2 generation

qiaouchicago:
https://jira.opensciencedatacloud.org/browse/DEV-1672

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -12,6 +12,7 @@ include:
 
 variables:
  BASE_CONTAINER_VERSION: "1.4.0"
+ PRE_COMMIT_DOCKER_IMAGE: docker.osdc.io/ncigdc/python3.8-builder:2.1.0
  DOCKER_BUILDKIT: 1
  PIP_EXTRA_INDEX_URL: https://nexus.osdc.io/repository/pypi-all/simple
  DOCKER_BUILD_OPTS: "--build-arg PIP_EXTRA_INDEX_URL=https://nexus.osdc.io/repository/pypi-all/simple"
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -41,7 +41,7 @@ release:
   rules:
     - when: never
 
-update_configs_and_reqs_for_gdcdatamodel2:
+update_reqs_for_gdcdatamodel2:
   image: docker.osdc.io/ncigdc/python36-builder:1.4.0
   stage: build
   script:
@@ -77,7 +77,7 @@ update_datamodels:
     - cat plaster.toml
     - bash plaster
     - git status
-    - rm setup.cfg dev-requirements.txt pyproject.toml
+    - rm setup.cfg dev-requirements.txt
   artifacts:
     paths:
       - gdcdatamodel2
@@ -104,8 +104,6 @@ release_datamodels_to_nexus:
     - twine check dist/*
     - twine upload dist/*
     - cp -v /tmp/pyproject.toml pyproject.toml
-    - ls -l pyproject.toml
-    - cat pyproject.toml
   variables:
     TWINE_USERNAME: ${TWINE_USER}
     TWINE_PASSWORD: ${TWINE_PASSWORD}
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -28,10 +28,33 @@ tox:
     matrix:
       - REPO_PY_VERSION: ["python36"]
 
+generate_new_dev_reqs_for_gdcdatamodel2:
+  image: docker.osdc.io/ncigdc/python36-builder:1.4.0
+  stage: build
+  script:
+    - !reference [.load_github_key, script]
+    - git clone git@github.com:NCI-GDC/gdcdatamodel2.git
+    - cd gdcdatamodel2
+    - pwd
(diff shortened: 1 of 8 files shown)


.gitlab-ci.yml
```diff
@@ -0,0 +1,147 @@
+---
+# To contribute improvements to CI/CD templates, please follow the Development guide at:
+# https://docs.gitlab.com/ee/development/cicd/templates.html
+# This specific template is located at:
+# https://gitlab.com/gitlab-org/gitlab/-/blob/master/lib/gitlab/ci/templates/Python.gitlab-ci.yml
```
kulgan, line 5: I don't think the comment about the location of this template is valid 🤔

.gitlab-ci.yml
```diff
@@ -0,0 +1,147 @@
...
+
+.python_versions:
+  parallel:
+    matrix:
+      - REPO_PY_VERSION: ["python36"]
+
+docker_build:
+  allow_failure: true
+  rules:
+    - when: never
```
kulgan, line 36: What does this stage do?

.gitlab-ci.yml
```diff
@@ -0,0 +1,147 @@
...
+
+release:
+  image: docker.osdc.io/ncigdc/python36-builder:1.4.0
+  allow_failure: true
+  rules:
+    - when: never
+
+update_reqs_for_gdcdatamodel2:
```
kulgan, line 44: I think this should use the `needs` option to depend on the main gdcdictionary tests completing successfully before triggering gdcdatamodel2 work. And can we rename the stage so its clearly different from the build step of gdcdictionary?

.gitlab-ci.yml
qiaouchicago: it suppose to build a docker image for gdcdictionary.

.gitlab-ci.yml
qiaouchicago: >I think this should use the needs option to depend on the main gdcdictionary tests completing successfully before triggering gdcdatamodel2 work
Screen Shot 2023-02-14 at 2 05 58 PM
It already does that. The later stage usually does not start until the earlier stage finish. The only exception is `code_quality` not sure why.

.gitlab-ci.yml
```diff
@@ -0,0 +1,151 @@
...
+include:
+  - project: nci-gdc/gitlab-templates
+    ref: master
+    file:
+      - templates/global/full.yaml
+      - templates/python/full.yaml
+
+variables:
+ BASE_CONTAINER_VERSION: "1.4.0"
+ PRE_COMMIT_DOCKER_IMAGE: docker.osdc.io/ncigdc/python3.8-builder:2.1.0
```
stilesj-uchicago, line 19: I think our pre-commit version needs to be 3.7 if we continue with the idea that we support the oldest supported version of python.

.gitlab-ci.yml
stilesj-uchicago: How is secret_detection different than the pre-commit step? Is it duplicating effort?

stilesj-uchicago (approved):

qiaouchicago:
>  How is secret_detection different than the pre-commit step? Is it duplicating effort?

It was the first to run. Another hook will modify the file and the .secret.baseline will be invalid. 
Moved it to last

qiaouchicago:
> How is secret_detection different than the pre-commit step? Is it duplicating effort?

I think it is, it is gitlab's implementation. 

https://docs.gitlab.com/ee/user/application_security/secret_detection/

.gitlab-ci.yml
qiaouchicago: I think it is, it is gitlab's implementation.

https://docs.gitlab.com/ee/user/application_security/secret_detection/

qiaouchicago: >And can we rename the stage so its clearly different from the build step of gdcdictionary?

Screen Shot 2023-02-14 at 3 39 29 PM

.gitlab-ci.yml
kulgan: What happens if you remove it, not sure why one will need a docker image of the gdcdictionary

.gitlab-ci.yml
kulgan: I think the difference is you want this stage to run only when the previous stage passes, not sure if that is handled automatically. For example, we don't want to build gdcdatamodel2 when gdcdictionary test is failing

.gitlab-ci.yml
```diff
@@ -0,0 +1,151 @@
...
+    - apt-get install -y python3-venv
+    - pip3 install -U setuptools_scm build twine setuptools
+    - cp -v pyproject.toml /tmp/pyproject.toml
+    - sed -i 's/local_scheme = "no-local-version"/local_scheme = "node-and-timestamp"/' pyproject.toml
+    - cat pyproject.toml
+    - python3 -m setuptools_scm
+    - python3 -m build
+    - twine check dist/*
+    - twine upload dist/*
+    - cp -v /tmp/pyproject.toml pyproject.toml
```
kulgan, line 110: As improvement for later, we can make use of the tox:release command defined the gdcdatamodel2's tox.ini

.gitlab-ci.yml
qiaouchicago: Screen Shot 2023-02-14 at 4 18 46 PM

qiaouchicago: I think the later stage won't run if a job in the previous stage failed.

kulgan (approved):

.gitlab-ci.yml
qiaouchicago: It will still run if I remove this part. It is defined in the template.

PR #964, opened 2023-05-01
Elion formatted

bullenc:
In preparation for the Galileo release, this PR is incorporating alphanum sorting of properties and enums in schema YAML files,  as well as implementing applicable enum value best practices (outlined in style guide here: https://uchicago.app.box.com/file/1106031699524) before updating nodes with  Galileo-specific updates. This is to reduce review burden for future PRs related to the Galileo release.


src/gdcdictionary/schemas/demographic.yaml
```diff
@@ -1180,16 +1180,16 @@ properties:
...
     enum:
-      - white
       - american indian or alaska native
-      - black or african american
       - asian
+      - black or african american
       - native hawaiian or other pacific islander
       - other
+      - unknown
+      - white
```
stilesj-uchicago, line 1189: white should bubble up to group the 'unknown' and 'Unknown' together. Add 'not allowed to collect' to the documentation regarding ordering.

src/gdcdictionary/schemas/diagnosis.yaml
```diff
@@ -19357,14 +19365,14 @@ properties:
       - G2
       - G3
       - G4
-      - GX
       - GB
+      - GX
       - High Grade
       - Intermediate Grade
       - Low Grade
+      - not reported
```
stilesj-uchicago, line 19373: Should probably group with the other `Not Reported` or update the documentation why this is belong in the alphabetical section

src/gdcdictionary/schemas/exposure.yaml
```diff
@@ -633,14 +633,14 @@ properties:
       - Smoking history not documented
       - Unknown
       - Not Reported
+      - Not Allowed To Collect
       - "1"
```
stilesj-uchicago, line 637: Numberic should be placed ahead of the 'Unknown, Not Reported, Not Allowed to Collect' set. (or the documentation is updated)

stilesj-uchicago (approved):
That is some eye-crossing work right there. lol.

src/gdcdictionary/schemas/demographic.yaml
wwysoc2: The "unknown" value has been deprecated, meaning that we are in the process of migrating it to "Unknown" and it should go away in the next release.  Good idea with adding "not allowed to collect" to the document.

src/gdcdictionary/schemas/follow_up.yaml
```diff
@@ -1554,34 +1554,34 @@ properties:
       - Gray
       - Green
       - Hazel
-      - Red & Violet
       - Other
+      - Red & Violet
```
wwysoc2, line 1558: Would "Red & Violet" go above "Other"?

bullenc: We don't have 'Other' indicated as a non-informative terms such as 'Unknown', 'Not Reported' etc. but it does make sense that it isn't really informative. I can add it to the docs and update the TSV2YAML to sort it towards the bottom of enum lists.

src/gdcdictionary/schemas/exposure.yaml
bullenc: I agree, the property these enums is attached to is excused from the auto sorting since it has frequency text terms and I overlooked it when reviewing the changes for a manual fix. I will update it.

src/gdcdictionary/schemas/diagnosis.yaml
bullenc: Added fix to script to group lowercase "not reported" with title case "Not Reported"

src/gdcdictionary/schemas/demographic.yaml
bullenc: Resolved ordering of enums, added 'not allowed to collect' sort guideline to the documentation.

wwysoc2 (approved):
With the changes suggested by @stilesj-uchicago , this looks good to me

PR #967, opened 2023-05-04
Galileo - MolecularTest

wwysoc2:
https://jira.opensciencedatacloud.org/browse/DICT-311

### DIFF
--- a/src/gdcdictionary/schemas/molecular_test.yaml
+++ b/src/gdcdictionary/schemas/molecular_test.yaml
@@ -7113,8 +7113,8 @@ properties:
     type: string
 
   staining_intensity_scale:
-    description: >-
-      The scale used by the immunohistochemistry staining rating.
+    $ref:
+      - "_terms.yaml#/staining_intensity_scale/common"
     enum:
       - 3 Point Scale
       - 4 Point Scale


src/gdcdictionary/schemas/molecular_test.yaml
```diff
@@ -7098,9 +7247,22 @@ properties:
       - "_terms.yaml#/test_value_range/common"
     enum:
       - 0-25
+      - 1–49%
+      - 10–19%
+      - 20–29%
       - 26-50
```
wwysoc2, line 7253: @bullenc , I'm going to group the whole numbers together and percentages together.  I foresee this being rare, so I don't think it would be all that necessary to add this to your formatting guide.

wteouchicago:
verified that the various values have been added. Not sure if intended, but I didn't see enumdef refs for these gene symbols:
- JARID2
- RPS15
- SETD1A
- HNRNPH1
- CTDP1
- SLC66A2
- CEP164

wteouchicago (approved):

clsackett (approved):

kulgan (approved):

sadhukd (approved):

src/gdcdictionary/schemas/molecular_test.yaml
```diff
@@ -7101,6 +7250,19 @@ properties:
...
       - 51-75
       - 76-100
```
stilesj-uchicago, line 7252: Are these meant to be deprecated going forward?

src/gdcdictionary/schemas/molecular_test.yaml
bullenc: Sounds good I'll remove this prop from being sorted in the script

bullenc:
> verified that the various values have been added. Not sure if intended, but I didn't see enumdef refs for these gene symbols:
> 
>     * JARID2
>     * RPS15
>     * SETD1A
>     * HNRNPH1
>     * CTDP1
>     * SLC66A2
>     * CEP164

@wteouchicago Great question, these values do not have corresponding definitions in NCIt (https://ncit.nci.nih.gov) so we have left them blank for refs and reached out to the NCIt team to see if these values can be added to the thesaurus in the future.

src/gdcdictionary/schemas/molecular_test.yaml
bullenc: @stilesj-uchicago Great q, this property actually captures a lot of different values from different questions on TCGA enrollment forms (https://www.nationwidechildrens.org/research/areas-of-research/biopathology-center/nci-ccg-project-team/the-cancer-genome-atlas/tcga-forms-and-documents) as well as other forms from other studies as well. It's sort of a catch all. 

As IMS is migrating clinical data from those forms, the values here will track the specific value ranges that were used for that specific study. So to answer, these ranges will probably not be deprecated, since we may not know the specific value for certain clinical data and where it would fall in a more specific range (e.g. would a value instance of 0-25 would be migrated to 0-10, 10-19 or 20-29?) We can probably bring up in data dict meeting to see how feasible this would be if we do have specific values if at all.

PR #968, opened 2023-05-04
Galileo - Treatment

wwysoc2:
https://jira.opensciencedatacloud.org/browse/DICT-312

### DIFF
--- a/src/gdcdictionary/schemas/treatment.yaml
+++ b/src/gdcdictionary/schemas/treatment.yaml
@@ -40,6 +40,7 @@ uniqueKeys:
 
 deprecated:
   - days_to_treatment
+  - treatment_anatomic_site
   - treatment_arm
 
 properties:
--- a/src/gdcdictionary/schemas/treatment.yaml
+++ b/src/gdcdictionary/schemas/treatment.yaml
@@ -289,6 +289,7 @@ properties:
       - NWTS-5
       - Not Reported
       - OSTEO 2006
+      - P9407
       - P9462
       - P9641
       - P9754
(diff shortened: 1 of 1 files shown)


src/gdcdictionary/schemas/treatment.yaml
```diff
@@ -6764,309 +6765,312 @@ properties:
...
-      - Vein
-      - Venous
-      - Vertebra
-      - Vulva
-      - White Blood Cells
-      - Wrist
-      - Other
-      - Unknown
-      - Not Reported
+    type: array
```
philtom-ctds, line 6768: Did this property's type change from a single enum to an array of enums?  if so, that's a breaking change.

wwysoc2: Good catch, will update node and ticket to deprecate/migrate singular to an array plurality (treatment_anatomic_sites)

src/gdcdictionary/schemas/treatment.yaml
```diff
@@ -7105,6 +7107,356 @@ properties:
         $ref:
           - "_terms_enum.yaml#/not_reported/common"
 
+  treatment_anatomic_sites:
+    $ref:
+      - "_terms.yaml#/treatment_anatomic_site/common"
```
wteouchicago, line 7112: should the ref be plural too? I mean "_terms.yaml#/treatment_anatomic_site**s**/common"

wwysoc2: I looked into this earlier.  Usually for plurals, the ref is identical to the singular, minus the plural title, so I decided to leave this one singular.  

We could add the plural ref to _terms.yaml, but I vote we add it after we merge the other changes, since there are so many moving parts here, all of which point at _terms.yaml.

wteouchicago: sgtm!

clsackett (approved):

sadhukd (approved):

src/gdcdictionary/schemas/treatment.yaml
```diff
@@ -7105,6 +7107,356 @@ properties:
...
+        - Venous
+        - Vertebra
+        - Vulva
+        - White Blood Cells
+        - Wrist
+        - Other
+        - Unknown
+        - Not Reported
+    enumDef:
+      Bone Marrow:
```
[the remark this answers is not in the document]
wwysoc2: This is usually (maybe always) a matter of whether or not each enum has been defined by an external source.

PR #969, opened 2023-05-04
Galileo - PathologyDetails

wwysoc2:
https://jira.opensciencedatacloud.org/browse/DICT-262

### DIFF
--- a/src/gdcdictionary/schemas/pathology_detail.yaml
+++ b/src/gdcdictionary/schemas/pathology_detail.yaml
@@ -580,8 +580,8 @@ properties:
           - "_terms_enum.yaml#/not_reported/common"
 
   measurement_type:
-    description: >-
-      The method used to measure tumor size.
+    $ref:
+      - "_terms.yaml#/measurement_type/common"
     enum:
       - Pathologic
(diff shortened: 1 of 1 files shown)


src/gdcdictionary/schemas/pathology_detail.yaml
```diff
@@ -860,12 +982,47 @@ properties:
         - Base
         - Middle
 
+  tumor_measurement_method:
+    description: >-
+      The method used to measure tumor size.
+    enum:
+      - Echographic
+      - Pathologic
```
wteouchicago, line 990: do we need enumdef refs for these?

wteouchicago: bill let me know this depends on whether we can find a def in an external database

wteouchicago (approved):

sadhukd (approved):
Looks good to me. Just for properties like `measurement_type` for example, we have a description here and is also in '_terms.yaml'.

clsackett (approved):

stilesj-uchicago (approved):

PR #970, opened 2023-05-04
Galileo - Demographic

wwysoc2:
https://jira.opensciencedatacloud.org/browse/DICT-263

### DIFF
--- a/src/gdcdictionary/schemas/demographic.yaml
+++ b/src/gdcdictionary/schemas/demographic.yaml
@@ -141,6 +141,940 @@ properties:
         $ref:
           - "_terms_enum.yaml#/not_reported/common"
 
+  country_of_birth:
+    description: >-
+      The name of the country in which the patient is born.
+    enum:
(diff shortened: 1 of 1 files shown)


src/gdcdictionary/schemas/demographic.yaml
```diff
@@ -141,6 +141,940 @@ properties:
         $ref:
           - "_terms_enum.yaml#/not_reported/common"
 
+  country_of_birth:
```
philtom-ctds, line 144: This property seems like it's going to have issues with history and time. For example, there are people who were born in countries that no longer exist, such as Yugoslavia or the USSR. Using only countries that exist today suggests that we're going to migrate this property over time as countries come in and out of existence.  

Anyway, aside from which enums to include, the real problem here probably belongs to groups like IMS, no us.

sadhukd (approved):
Looks good to me.

clsackett (approved):

PR #972, opened 2023-05-04
Galileo - FollowUp

wwysoc2:
https://jira.opensciencedatacloud.org/browse/DICT-265

### DIFF
--- a/src/gdcdictionary/schemas/follow_up.yaml
+++ b/src/gdcdictionary/schemas/follow_up.yaml
@@ -1108,6 +1108,7 @@ properties:
         - Peripheral Vascular Disease
         - Peutz-Jeghers Disease
         - Pneumocystis Pneumonia
+        - Pneumonia, NOS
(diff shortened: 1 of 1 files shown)


clsackett (approved):

sadhukd (approved):

src/gdcdictionary/schemas/follow_up.yaml
```diff
@@ -1597,6 +1608,7 @@ properties:
       - Induction Death
       - Induction Failure
       - None
+      - Presented With Mets
```
stilesj-uchicago, line 1611: A bio question, what is presented with mets? Google was useless unless I want to learn about baseball.

stilesj-uchicago (approved):

src/gdcdictionary/schemas/follow_up.yaml
wwysoc2: Mets = Metastases, I never thought of how hard that would be to Google!

PR #979, opened 2023-05-04
Galileo - AlignedReads

wwysoc2:
https://jira.opensciencedatacloud.org/browse/DICT-307

### DIFF
--- a/src/gdcdictionary/schemas/aligned_reads.yaml
+++ b/src/gdcdictionary/schemas/aligned_reads.yaml
@@ -223,8 +223,8 @@ properties:
     type: number
 
   wgs_coverage:
-    description: >-
-      Range of coverage values for WGS aligned reads.
+    $ref:
+      - "_terms.yaml#/wgs_coverage/common"
     enum:
       - 0x–10x
       - 10x–25x
--- a/src/gdcdictionary/schemas/aligned_reads.yaml
+++ b/src/gdcdictionary/schemas/aligned_reads.yaml
@@ -222,6 +222,24 @@ properties:
       - "_terms.yaml#/tumor_purity/common"
     type: number
 
+  wgs_coverage:
+    description: >-
+      Range of coverage values for WGS aligned reads.
+    enum:
+      - 0x–10x
+      - 10x–25x
+      - 25x–150x
+      - 150x+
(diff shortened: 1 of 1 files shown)


src/gdcdictionary/schemas/aligned_reads.yaml
```diff
@@ -222,6 +222,24 @@ properties:
       - "_terms.yaml#/tumor_purity/common"
     type: number
 
+  wgs_coverage:
+    description: >-
```
wteouchicago, line 226: _terms file has wgs_coverage, ref?

wwysoc2: Fixed to point to ref.  Will do for other nodes too.

wteouchicago (approved):

clsackett (approved):

sadhukd (approved):

kulgan (approved):

src/gdcdictionary/schemas/aligned_reads.yaml
```diff
@@ -222,6 +222,24 @@ properties:
...
     type: number
 
+  wgs_coverage:
+    $ref:
+      - "_terms.yaml#/wgs_coverage/common"
+    enum:
+      - 0x–10x
+      - 10x–25x
+      - 25x–150x
+      - 150x+
```
philtom-ctds, line 232: I'm suggesting any changes here.  I just want to clarify the intentions, as developers will need to know later.

Are these ranges an inclusive lower bound and exclusive upper bound?  for example:

- 0x-10x == 0, 1, 2, 3, 4, 5, 6, 7, 8, 9
- 10x - 25x == 10, 11, ..., 23, 24
- 25x - 150x == 25, 26, ... 148, 149
- 150x+ == 150 and higher

wwysoc2: Since most of these values will have many decimal places, it's very rare that it will ever fall directly on one of the bounds that are shared between categories.  However, in the spirit of safety and consistency, we will add the following sentence to the description of this value in the _terms file:  "Each range is non-inclusive to the lower bound."    

We decided to go in this direction because the coverage will never be 0x.

PR #981, opened 2023-05-09
added new preservation_method enum 'EDTA'

bullenc:
Adding new enum for sample.preservation_method, 'EDTA'

### DIFF
--- a/src/gdcdictionary/schemas/sample.yaml
+++ b/src/gdcdictionary/schemas/sample.yaml
@@ -747,6 +747,7 @@ properties:
       - FFPE
       - Fresh
       - Frozen
+      - EDTA
       - OCT
       - Snap Frozen
       - Unknown


src/gdcdictionary/schemas/sample.yaml
```diff
@@ -747,6 +747,7 @@ properties:
       - FFPE
       - Fresh
       - Frozen
+      - EDTA
```
mpsolano: Should an enumDef be added for this?

bullenc: None of the other specific values have one, but it does look a [term](https://ncit.nci.nih.gov/ncitbrowser/ConceptReport.jsp?dictionary=NCI_Thesaurus&ns=ncit&code=C61742) does exist; we would have to update the terms YAML first to not break anything with this PR though.

mpsolano: Understood, thank you all for the feedback!

mpsolano (approved):

stilesj-uchicago (approved):

PR #984, opened 2023-05-12
DEV-1849: Fix unicode chars in schema files.

wrkhenddher:


.pre-commit-config.yaml
```diff
@@ -1,23 +1,64 @@
...
+        pass_filenames: false
+        entry: |
+          sh -c "
+          # echo -n '–' | hexdump -C # > e2 80 93
+          LC_ALL=C
+          for f in $(ls src/gdcdictionary/schemas/*.yaml); do
+            # echo 'Converting ' $f
+            sed -i '' 's/\xe2\x80\x93/-/g' $f
+          done
+          "
```
philtom-ctds, line 41: 1. This isn't going to work for everyone, such as Windows users. 
2. If we are going to enforce ascii, we could use [require-ascii](https://github.com/jumanjihouse/pre-commit-hooks#require-ascii). It's catch much more than just the hyphen. We'll need to exclude [these files](https://github.com/NCI-GDC/gdcdictionary/blob/4068e3d6dc287a4251f64670cb89aa2469407b5b/src/gdcdictionary/python/__init__.py#L34-L38), as described in [this comment](https://github.com/NCI-GDC/gdcdictionary/blob/4068e3d6dc287a4251f64670cb89aa2469407b5b/src/gdcdictionary/python/__init__.py#L66-L67).

.pre-commit-config.yaml
```diff
@@ -1,23 +1,64 @@
...
+  - repo: local
+    hooks:
+      - id: replacer
+        # More chars can be added
+        name: Replace unicode dashes with ascii dashes.
+        language: system
+        pass_filenames: false
+        entry: |
+          sh -c "
+          # echo -n '–' | hexdump -C # > e2 80 93
```
philtom-ctds, line 35: this is a great comment 👍

.pre-commit-config.yaml
```diff
@@ -1,23 +1,64 @@
...
+
+  - repo: https://github.com/jumanjihouse/pre-commit-hook-yamlfmt
+    rev: 0.2.1
+    hooks:
+      - id: yamlfmt
+        args: [--mapping, '2', --sequence, '4', --offset, '2', --width, '130']
+        # TODO: Remove stages so it always run. Do it once all yamls are properly formatted.
+        stages: [manual]
+
+  - repo: https://github.com/adrienverge/yamllint.git
```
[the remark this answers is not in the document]
wrkhenddher: Even when we neither distribute the tool, nor modify it, nor include it in our distribution?

(wouldn't using this tool be equivalent to using any other gnu-tool in our build system?)

.pre-commit-config.yaml
wrkhenddher: > 1. This isn't going to work for everyone, such as Windows users.

Good point. Should it be a deal breaker?

> 2. If we are going to enforce ascii, we could use [require-ascii](https://github.com/jumanjihouse/pre-commit-hooks#require-ascii). It's catch much more than just the hyphen. We'll need to exclude [these files](https://github.com/NCI-GDC/gdcdictionary/blob/4068e3d6dc287a4251f64670cb89aa2469407b5b/src/gdcdictionary/python/__init__.py#L34-L38), as described in [this comment](https://github.com/NCI-GDC/gdcdictionary/blob/4068e3d6dc287a4251f64670cb89aa2469407b5b/src/gdcdictionary/python/__init__.py#L66-L67).

The catch is that unicode chars are allowed. Depending on the version of the yaml file (which we do NOT explicitly define) and the version of the parser (somehow loose atm), one should either quote them or do whatever applies. 

When I first saw the failure and the traceback, I was inclined to remove the whole "ascii" enforcement and instead specify yaml versions and straighten up all the versions (yaml files and parser). 

I think we should pursue that end. It might take some bit effort but as long as we can diff the current json schema with the new one and make sure the match, we should be ok. IMHO.

PR #966, opened 2023-05-03
Galileo DICT-314 updates to terms and terms_enum YAMLs

bullenc:
Updates to terms and enum terms YAML files; changes include additions of new enums and property descriptions/references as well as update to URLs for CDE codes to caDSR II browser.

PR #986, opened 2023-05-16
Schema issues?

wrkhenddher:


src/gdcdictionary/schemas/tag.yaml
```diff
@@ -119,4 +119,4 @@ properties:
     enumDef:
       exon:
         $ref:
-          - "_terms_enum.yaml#/exon/common"
+          - "_terms.yaml#/exon/common"
```
philtom-ctds, line 122: _terms_enum is the correct file.  but it looks like the `exon` entry was moved in a recent [PR](https://github.com/NCI-GDC/gdcdictionary/pull/966/files) and the nested structure was changed.

The [new entry](https://github.com/NCI-GDC/gdcdictionary/pull/966/files#diff-4f9b039beb68e976da2792bc3e1bdb1ed66ad987e66974aa9c3c05f4f8fe7792R102337-R102350) vs [the old entry](https://github.com/NCI-GDC/gdcdictionary/pull/966/files#diff-4f9b039beb68e976da2792bc3e1bdb1ed66ad987e66974aa9c3c05f4f8fe7792L44180-L44193).  @clsackett : was this change intentional or was it a mistake?

bullenc: I made this change, there was conflict with having two 2 instances of exon/common being referenced between the _terms_enum.yaml and _terms.yaml files with our internal scripts. Per @wwysoc2 this enumDef isn't really necessary anymore since tags schema are not used in active portal, and its the only one that uses this enumDef. It may make more sense to just remove this ref from the tags.yaml file if it is causing issues.

wrkhenddher: https://github.com/NCI-GDC/gdcdictionary/pull/988 should take care of this typo.
Will close this PR.

PR #987, opened 2023-05-16
Fix commands property in tox.ini

wrkhenddher:
- According to [tox docs], the exit code of any command preceeded by a dash is ignored.
[tox docs]: https://tox.wiki/en/latest/config.html#commands
- Unless 'env' is in 'allowlist_externals', it will not run until completion.


kulgan (approved):
lgtm

BenX00 (approved):

PR #995, opened 2023-06-07
DEV 1740 move release gdcdatamodel2 to gdcdatamodel2

qiaouchicago:
this PR depends on the following change in gdcdatamodel2: https://github.com/NCI-GDC/gdcdatamodel2/pull/10


mpsolano (approved):
I believe this PR depends on the following change in gdcdatamodel2: https://github.com/NCI-GDC/gdcdatamodel2/pull/10. I would probably add this detail to the description of the PR so reviewers are aware of that change as well.

PR #1011, opened 2023-10-04
Inoue pathology detail dict 345

bullenc:


src/gdcdictionary/schemas/pathology_detail.yaml
```diff
@@ -236,6 +236,14 @@ properties:
         $ref:
           - "_terms_enum.yaml#/not_reported/common"
 
+  days_to_pathology_detail:
+    description: >-
```
wteouchicago, line 240: should we put the description in _terms file? ditto `micrometastasis_present` below

wteouchicago (approved):

src/gdcdictionary/schemas/pathology_detail.yaml
bullenc: Yeah we need to flesh out a process for what to do in terms definition adding when we do not have a code in a terms database (NCIt, caDSR etc) but still want to offer a description, especially for properties. I think for now we're ok but if/when the EVS team hopefully generates a code for this we can formally add to terms file with associated definition ID.

clsackett (approved):

src/gdcdictionary/schemas/pathology_detail.yaml
```diff
@@ -350,6 +400,25 @@ properties:
...
+  histologic_progression_type:
+    $ref:
+      - "_terms.yaml#/histologic_progression_type/common"
+    enum:
+      - Anaplastic
+      - Poorly differentiated
+      - Unknown
+      - Not Reported
+    enumDef:
+      Poorly differentiated:
```
wrkhenddher, line 412: Isn't this missing `Anaplastic` ?

src/gdcdictionary/schemas/pathology_detail.yaml
```diff
@@ -982,13 +1080,6 @@ properties:
         - Base
         - Middle
 
-  tumor_measurement_method:
-    $ref:
-      - "_terms.yaml#/tumor_measurement_method/common"
-    enum:
-      - Echographic
-      - Pathologic
-
```
wrkhenddher, line 991: Are these deprecated?
or just removed outright ?

wrkhenddher: and what about it in `_terms.yaml`?

wrkhenddher (approved):

src/gdcdictionary/schemas/pathology_detail.yaml
bullenc: We are removing these outright- they were not used and are redundant with another property.

PR #1014, opened 2023-10-04
update sample  node DICT-349 Inoue

bullenc:


wteouchicago (approved):

clsackett (approved):

qiaouchicago (approved):

src/gdcdictionary/schemas/sample.yaml
```diff
@@ -195,6 +195,7 @@ properties:
       - Hard Palate
       - Head & Neck
       - Head - Face Or Neck, Nos
```
wrkhenddher, line 197: And this one stays here?

wrkhenddher (approved):

src/gdcdictionary/schemas/sample.yaml
bullenc: Correct, it stays but an entry is created under 'deprecated_enum' to indicate it has been deprecated from use.

wwysoc2 (approved):
lgtm!

PR #1022, opened 2024-04-23
updates to _definitions.yaml leeuwenhoek

bullenc:
See https://gdc-ctds.atlassian.net/browse/DICT-455 for expected changes

### DIFF
--- a/src/gdcdictionary/schemas/_definitions.yaml
+++ b/src/gdcdictionary/schemas/_definitions.yaml
@@ -81,11 +81,11 @@ datetime:
     - type: 'null'
 
 file_name:
-  $ref: "_terms.yaml#/file_name"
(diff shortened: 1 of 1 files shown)


wwysoc2 (approved):
lgtm!

stilesj-uchicago (approved):
What is up with the common construct? Is that to support special cases?

bullenc:
> What is up with the common construct? Is that to support special cases?

@stilesj-uchicago Right, sometimes we have terms that have multiple definitions, e.g. staging of cancers I-IV can have different definitions for different cancers. `common` represents general definition of a term, and more specific nodes-property links can be substituted for `common` when we run into multiple definitions for a term and to differentiate them.

PR #1026, opened 2024-04-24
new node submitted_expression_array leeuwenhoek

bullenc:
https://gdc-ctds.atlassian.net/browse/DICT-423

### DIFF
--- a/src/gdcdictionary/schemas/annotation.yaml
+++ b/src/gdcdictionary/schemas/annotation.yaml
@@ -310,6 +310,12 @@ links:
         target_type: submitted_aligned_reads
         multiplicity: many_to_many
         required: false
+      - name: submitted_expression_arrays
+        backref: annotations
+        label: annotates
+        target_type: submitted_expression_array
+        multiplicity: many_to_many
+        required: false
       - name: submitted_genomic_profiles
         backref: annotations
         label: annotates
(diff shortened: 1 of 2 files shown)


wteouchicago:
Looks ok to me -- I compared to another submitted array as a sanity check.

wteouchicago (approved):

wwysoc2 (changes requested):
I saw that gitlab failed, I can't get into that, but my guess is that we need to add the annotation link.

ebusygina (approved):

src/gdcdictionary/schemas/annotation.yaml
```diff
@@ -310,6 +310,12 @@ links:
         target_type: submitted_aligned_reads
         multiplicity: many_to_many
         required: false
+      - name: submitted_expression_arrays
```
stilesj-uchicago, line 313: Should this name be submitted_expressions? (the plural indicates a collection of values). Unless 'expression array' is the term used in the vocabulary. I was thinking 'array' was indicating a software type.

stilesj-uchicago (approved):
lgtm

src/gdcdictionary/schemas/annotation.yaml
bullenc: I think `submitted_expression_arrays` is correct, since this would be the backref from annotation to find the edges it is connected to, i.e. for case it is `cases`:

```
In [24]: g.nodes(Annotation).first().cases
Out[24]: [<Case(b5622ca2-8f51-453e-b411-b2ac045bb04a)>]
```

wwysoc2 (approved):

PR #1042, opened 2024-04-24
updates to masked_somatic_mutation node leeuwenhoek

bullenc:
https://gdc-ctds.atlassian.net/browse/DICT-441 + alphabetizing

### DIFF
--- a/src/gdcdictionary/schemas/masked_somatic_mutation.yaml
+++ b/src/gdcdictionary/schemas/masked_somatic_mutation.yaml
@@ -73,18 +73,18 @@ properties:
     enum:
       - Simple Nucleotide Variation
 
-  data_type:
-    $ref:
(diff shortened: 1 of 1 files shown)


ebusygina (approved):

wwysoc2 (changes requested):
I was just thinking, these are the type of nodes that were tumor-only and supposed to be controlled, right?  Should we add an enum the same as we did for aggregated somatic mutation?
EDIT - NVM, that's on the workflow, we have it already.  Disregard this.

wwysoc2 (approved):
👍

wteouchicago (approved):

stilesj-uchicago (approved):

stilesj-uchicago (approved):
lgtm

PR #1050, opened 2024-04-26
DICT-445 leeuwenhoek treatment updates

bullenc:
https://gdc-ctds.atlassian.net/browse/DICT-445

### DIFF
--- a/src/gdcdictionary/schemas/treatment.yaml
+++ b/src/gdcdictionary/schemas/treatment.yaml
@@ -114,6 +114,8 @@ properties:
       - Glucocorticoid
       - Growth factor
       - PARP inhibitor
+      - Unknown
+      - Not Reported
 
   embolic_agent:
     $ref:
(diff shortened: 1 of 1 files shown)


stilesj-uchicago (approved):
lgtm

wwysoc2 (approved):
👍

src/gdcdictionary/schemas/treatment.yaml
```diff
@@ -172,6 +176,37 @@ properties:
...
+      - mg/m2/wk
+      - mg/mL
+      - mg/wk
+      - mIU
+      - mL
+      - ug
+      - ug/m2
+      - Wafer
+      - Unknown
+      - Not Reported
```
wteouchicago, line 208: should we update the DICT ticket to indicate that these two are to be added?

bullenc: Yes, thank you

wteouchicago (approved):

PR #1051, opened 2024-04-26
DICT-446 leeuwenhoek other_clinical_attribute updates

bullenc:
https://gdc-ctds.atlassian.net/browse/DICT-446

### DIFF
--- a/src/gdcdictionary/schemas/other_clinical_attribute.yaml
+++ b/src/gdcdictionary/schemas/other_clinical_attribute.yaml
@@ -47,6 +47,10 @@ uniqueKeys:
   - [id]
   - [project_id, submitter_id]
 
+deprecated:
+  - pregnancy_count
+  - viral_hepatitis_serologies
+
 properties:
 
   $ref:
@@ -363,6 +367,13 @@ properties:
     maximum: 32872
     minimum: -32872
 
+  days_to_risk_factor:
+    $ref:
+      - "_terms.yaml#/days_to_risk_factor/common"
+    type: integer
+    maximum: 32872
+    minimum: -32872
+
   diabetes_treatment_type:
     $ref:
       - "_terms.yaml#/diabetes_treatment_type/common"
@@ -423,6 +434,7 @@ properties:
       - Hazel
       - Red & Violet
       - Other
+      - Unknown
       - Not Reported
     enumDef:
       Other:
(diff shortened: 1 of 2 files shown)


src/gdcdictionary/schemas/other_clinical_attribute.yaml
```diff
@@ -394,7 +405,6 @@ properties:
     $ref:
       - "_terms.yaml#/dlco_ref_predictive_percent/common"
     type: number
-    maximum: 100
```
wteouchicago, line 397: is there a reason why we are removing the maxes? can percent go over 100?

wwysoc2: We found that out yesterday from Megan, apparently these are lung capacity tests and they can go over 100

wteouchicago (approved):

wwysoc2 (dismissed):
I think hormonal_replacement_therapy_status was left out.  The ticket says "add enum" but I think it's a property we need to add.

Also, did we figure out the pregnancy_count values that were already present?

src/gdcdictionary/schemas/other_clinical_attribute.yaml
stilesj-uchicago, line 1: Am I missing the 'risk_factors','timepoint_category','hormonal_replacement_therapy_status','immunosuppressive_treatment_type', etc enums  that were to be added?

Note: timepoint_category is listed twice in the ticket.

bullenc: I see the enums for the properties you listed except for hormonal_replacement_therapy_status. I merged the two separate enum additions in the table in the ticket (`Unknown` and `Sample Procurement`).

bullenc:
> I think hormonal_replacement_therapy_status was left out. The ticket says "add enum" but I think it's a property we need to add.

Ok I will create a new _terms.yaml ticket and circle back to this PR once we include the new property in terms.yaml file

> Also, did we figure out the pregnancy_count values that were already present?

Yes, the only submitted values were for test/QA projects, pretty sure we don't need to migrate those.

src/gdcdictionary/schemas/other_clinical_attribute.yaml
stilesj-uchicago: I must've gone cross-eyed. I see them now, except for hormonal_replacement_therapy_status

stilesj-uchicago (approved):

PR #1052, opened 2024-04-26
DICT-447 leeuwenhoek exposure  updates

bullenc:
https://gdc-ctds.atlassian.net/browse/DICT-447

### DIFF
--- a/src/gdcdictionary/schemas/exposure.yaml
+++ b/src/gdcdictionary/schemas/exposure.yaml
@@ -83,6 +83,8 @@ properties:
     enum:
       - Daily Drinker
       - Weekly Drinker
+      - Not Reported
+      - Unknown
 
   alcohol_history:
     $ref:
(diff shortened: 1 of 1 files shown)


src/gdcdictionary/schemas/exposure.yaml
```diff
@@ -247,6 +252,7 @@ properties:
       - "Yes"
       - "No"
       - Unknown
+      - Not Reported
```
wwysoc2: Just realized this was deprecated, but not a big deal if this is kept in, we just need to remove it from the release notes.

wwysoc2 (approved):

wteouchicago (approved):

src/gdcdictionary/schemas/exposure.yaml
bullenc: Removed it to be consistent

stilesj-uchicago (approved):
lgtm

PR #1054, opened 2024-04-26
DICT-448 leeuwenhoek follow_up updates

bullenc:
https://gdc-ctds.atlassian.net/browse/DICT-448

### DIFF
--- a/src/gdcdictionary/schemas/follow_up.yaml
+++ b/src/gdcdictionary/schemas/follow_up.yaml
@@ -61,6 +61,7 @@ deprecated:
   - comorbidity_method_of_diagnosis
   - comorbidities
   - days_to_comorbidity
+  - days_to_risk_factor
   - diabetes_treatment_type
   - dlco_ref_predictive_percent
   - eye_color
@@ -208,6 +209,7 @@ properties:
       - Carbon Monoxide Diffusing Capacity Decreased
       - Cardiac Arrest
       - Cardiac Disorders - Other
+      - Cardiac Ischemia
       - Cardiac Troponin I Increased
       - Cardiac Troponin T Increased
       - Cataract
@@ -384,6 +386,7 @@ properties:
       - GGT Increased
       - Gingival Pain
       - Glaucoma
+      - Glomerular Filtration Rate Decreased
       - Glossopharyngeal Nerve Disorder
       - Glucose Intolerance
       - Growth Accelerated
@@ -504,6 +507,7 @@ properties:
       - Jejunal Ulcer
       - Joint Effusion
       - Joint Infection
+      - Joint Pain
       - Joint Range of Motion Decreased
       - Joint Range of Motion Decreased Cervical Spine
       - Joint Range of Motion Decreased Lumbar Spine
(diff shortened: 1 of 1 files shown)


src/gdcdictionary/schemas/follow_up.yaml
```diff
@@ -2953,11 +2988,26 @@ properties:
...
+      - Not Reported
+
+  treatment_emergent_adverse_event:
+    $ref:
+      - "_terms.yaml#/treatment_emergent_adverse_event/common"
+    enum:
+      - Definitely
+      - Possibly
+      - Probably
+      - Unlikely
```
stilesj-uchicago, line 3008: I don't see a difference between Possibly and Unlikely. The seems like 'half-full' and 'half-empty'. Are there metrics related to these terms?

src/gdcdictionary/schemas/follow_up.yaml
stilesj-uchicago, line 1: ~I was looking for a new `adverse_event`,`first_event`, enum~
first_event has an additional 'Unknown' not documented in the associated ticket.

stilesj-uchicago (approved):

src/gdcdictionary/schemas/follow_up.yaml
bullenc: I'll defer to @tlicht3 , IMS gets these values from NCI studies and these might be 1:1 to what needs to be populated from the CTSP-KIRC study.

src/gdcdictionary/schemas/follow_up.yaml
bullenc: There was an additional row for `first_event` in the table that had `Unknown`, consolidated the two rows

src/gdcdictionary/schemas/follow_up.yaml
wwysoc2: My guess is that these are the selectable options on some doctors' form.  Based on the CTSP-KIRC trial information, Definitely, Probably, and Possibly fell into one category, and the Unlikely and Unrelated fell into another.

src/gdcdictionary/schemas/follow_up.yaml
```diff
@@ -2953,11 +2988,26 @@ properties:
...
+      - Unrelated
+      - Unknown
       - Not Reported
```
wteouchicago, line 3011: confused about why this Not Reported was already here. I think it was already there previously for timepoint category and git is just confused?
^^ if so, should we remove "Not Reported" from timepoint_category in the jira?

wteouchicago (approved):

src/gdcdictionary/schemas/follow_up.yaml
bullenc: Yeah I saw that too, I think its just the way that git rendered the diff for changes. I think the timepoint_category enum for Not Reported should remain so we are consistent that all enum type props have `Unknown` and `Not Reported`

wteouchicago: oh, i mean if timepoint_category already had "Not Reported", then we don't need to claim we added it for Leeuvenhoek in the jira ticket. But not wedded to this in any way.

bullenc: Ok yes, I see what you mean now. timepoint_category already had `Not Reported` as an enum in the dictionary, removed it as a proposed addition from the ticket.

PR #1062, opened 2024-05-15
DEV-2738: add json schema validator

kulgan:
Add validator module with functions that validates a user supplied json document. Minor refactoring on the gdcdictionary module by doing the following:
* Rename python/__init__.py to core.py
* Replace path traversal with importlib-resources files which is more consistent with loading resources in python
* Refactor tests to use importlib-resources as well

### DIFF
--- a/src/gdcdictionary/__init__.py
+++ b/src/gdcdictionary/__init__.py
@@ -1,7 +1,17 @@
-import os
+from gdcdictionary.core import GDCDictionary, gdcdictionary, get_schema_directory
+from gdcdictionary.validators import (
+    SchemaValidator,
+    SchemaValidationError,
+    validate,
+    validate_instances,
+)
 
-from gdcdictionary.python import GDCDictionary, gdcdictionary
-
-
-
-ROOT_DIR = os.path.dirname(os.path.abspath(__file__))
+__all__ = [
+    "gdcdictionary",
+    "get_schema_directory",
+    "validate",
+    "validate_instances",
+    "GDCDictionary",
+    "SchemaValidator",
+    "SchemaValidationError",
+]
--- a/src/gdcdictionary/core.py
+++ b/src/gdcdictionary/core.py
@@ -7,7 +7,7 @@ try:
     from importlib.resources import files
 except ImportError:
     from importlib_resources import files
-from typing import Optional, List
+from typing import Optional, List, Tuple, Dict, Any
 
 from jsonschema import RefResolver
 
@@ -88,7 +88,7 @@ class GDCDictionary:
             logger.debug("To enable CSafeLoader install libyaml. Falling back to yaml.safe_load()")
             return yaml.safe_load(f)
 
-    def load_schemas_from_dir(self, directory: pathlib.Path):
+    def load_schemas_from_dir(self, directory: pathlib.Path) -> Tuple[Dict[str, Any], Dict[str, Any]]:
         """Returns all yamls and resolvers of those yamls from dir"""
 
         schemas, resolvers = {}, {}
@@ -101,7 +101,7 @@ class GDCDictionary:
 
         return schemas, resolvers
 
-    def load_directory(self, directory: pathlib.Path):
+    def load_directory(self, directory: pathlib.Path) -> None:
         """Load and resolve all schemas from directory"""
 
         yamls, resolvers = self.load_schemas_from_dir(directory)
--- a/src/gdcdictionary/core.py
+++ b/src/gdcdictionary/core.py
@@ -1,31 +1,39 @@
+import logging
+import pathlib
 from copy import deepcopy
 from collections import namedtuple
-from contextlib import contextmanager
+
+from importlib import resources
+from typing import Optional, List
+
 from jsonschema import RefResolver
 
-import glob
-import logging
-import os
 import yaml
 
-MOD_DIR = os.path.dirname(os.path.abspath(os.path.dirname(__file__)))
 
+logger = logging.getLogger(__name__)
 ResolverPair = namedtuple('ResolverPair', ['resolver', 'source'])
 
 
-@contextmanager
-def visit_directory(path):
-    """Perform contained actions with current working directory at
-    :param:``path``.  Always return to previous directory when done.
+def get_schema_directory(local_path: Optional[str] = None) -> pathlib.Path:
+    """Resolve the directory containing the schema definitions files.
 
+    Args:
+        local_path: custom location for the schema
+    Returns:
+        the schema directory as path object.
     """
+    if local_path:
+        path = pathlib.Path(local_path)
 
-    cdir = os.getcwd()
-    try:
-        os.chdir(path)
-        yield os.getcwd()
-    finally:
-        os.chdir(cdir)
+        if not path.exists():
+            raise IOError("Specified template directory '%s' does not exist", path)
+        return path
+
+        # use default embedded location
+    with resources.path("gdcdictionary", "schemas") as path:
+        logger.info(path)
+        return path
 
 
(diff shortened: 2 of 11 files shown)


src/gdcdictionary/__init__.py
```diff
@@ -1,7 +1,17 @@
...
+    validate,
+    validate_instances,
+)
 
-from gdcdictionary.python import GDCDictionary, gdcdictionary
-
-
-
-ROOT_DIR = os.path.dirname(os.path.abspath(__file__))
+__all__ = [
```
ProfOak, line 9: I haven't seen this before. Can you explain this a little?

src/gdcdictionary/core.py
```diff
@@ -71,29 +81,28 @@ def load_yaml(self, name):
...
+                    logger.error(f"Error in file: {name}")
                     raise e
             if yaml.__with_libyaml__:
                 return yaml.load(f, Loader=yaml.CSafeLoader)
-            self.logger.debug("To enable CSafeLoader install libyaml. Falling back to yaml.safe_load()")
+            logger.debug("To enable CSafeLoader install libyaml. Falling back to yaml.safe_load()")
             return yaml.safe_load(f)
 
-    def load_schemas_from_dir(self, directory):
+    def load_schemas_from_dir(self, directory: pathlib.Path):
```
mpsolano: It would be nice to include a type hint for the return here.

src/gdcdictionary/core.py
```diff
@@ -71,29 +81,28 @@ def load_yaml(self, name):
...
+            schema = self.load_yaml(path)
+            schemas[path.name] = schema
+            resolver = RefResolver(f'{path.name}#', schema)
+            resolvers[path.name] = ResolverPair(resolver, schema)
 
         return schemas, resolvers
 
-    def load_directory(self, directory):
-        """Load and reslove all schemas from directory"""
+    def load_directory(self, directory: pathlib.Path):
```
mpsolano: Just for consistency, it would be nice to include None as the return type.

src/gdcdictionary/validators.py
```diff
@@ -0,0 +1,155 @@
...
+    Properties:
+        name: name of specific dictionary schema
+        validator: json validator initialized with the schema
+    """
+
+    def __init__(self, name: str) -> None:
+        self.name = name
+        self.validator = Draft4Validator(gdcdictionary.schema[name])
+
+    def post_validate(self, violations: List[SchemaValidationError]) -> None:
```
mpsolano: Do we have specific plans to implement logic later for `post-validate` and `pre_validate`?

ProfOak (approved):

tests/test_validators.py
```diff
@@ -0,0 +1,79 @@
...
+        assert len(gdcdictionary.validate(doc)) == 0
+
+
+@pytest.mark.parametrize("path", get_all_paths("invalid"))
+def test_invalid_examples(path, schema):
+    with open(path) as f:
+        doc = json.load(f)
+        violations = gdcdictionary.validate(doc)
+        assert len(violations) > 0
+        print(violations)
```
mpsolano, line 40: Is the print still needed?

tests/test_validators.py
```diff
@@ -0,0 +1,79 @@
...
+    instances = []
+    for path in paths:
+        with open(path) as f:
+            doc = json.load(f)
+            instances.append(doc)
+    violations = gdcdictionary.validate_instances(instances)
+    assert len(violations) == 0
+
+
+def test_validate_instances__invalid_types():
```
mpsolano: To keep the focus of the tests narrow, an optional suggestion would be to split this into multiple tests (one for unknown types, one for missing required and one for extra fields).

mpsolano (approved):
Looks good!

src/gdcdictionary/__init__.py
kulgan: It's for specifying portion of code that the public should be using. Things not publicly exposed can be removed or refactored without notice, but these ones should remain the same and whenever they change we should give some sort of notice and probably deprecate first.

src/gdcdictionary/validators.py
kulgan: I am not aware of any specific plans. The existing validator does leave room for post validation work [here](https://github.com/NCI-GDC/gdcdatamodel/blob/09f7e13da37387a5bb364c0e21467cafb465cdc0/gdcdatamodel/validators/json_validators.py#L61). The idea itself is useful and would provide one way to do partial validation, where certain validation errors can be filter out

tests/test_validators.py
kulgan: Its ok to leave it for debug purposes

src/gdcdictionary/validators.py
mpsolano: Perhaps I'm misunderstanding the reason why both these methods are left unimplemented. I don't see that SchemaValidator is an ABC nor do I see it extended anywhere. I'm sure it's just a misunderstanding on my part but if you could provide more context on that I would appreciate it.

tests/test_validators.py
kulgan: I ended up leaving the test as is and added log for the violations. The reason is that, this test is essentially the same as `test_invalid_examples` but its testing using a bulk json. The violations (see snippet below) I asserted are a subset of the resulting errors from the three scenarios in the input. These are not all possible combination of violations, they are just the three in the sample input. Splitting this into three makes it look like these 3 are more important than others, which they are not and I cannot write tests for all the possible combinations.
``` python
 [
        SchemaValidationError(
            schema="",
            message="specified type: species is not in the current data model",
            keys=["type"],
        ),
        SchemaValidationError(
            schema="case",
            message="Key(s) ['python_version'] not a valid property for type 'case'",
            keys=["python_version"],
        ),
        SchemaValidationError(
            schema="case",
            message="'submitter_id' is a required property",
            keys=["submitter_id"],
        ),
        SchemaValidationError(
            schema="case",
            message="'disease_type' is a required property",
            keys=["disease_type"],
        ),
        SchemaValidationError(
            schema="case",
            message="'primary_site' is a required property",
            keys=["primary_site"],
        ),
        SchemaValidationError(
            schema="", message="'type' is a required property", keys=["type"]
        ),
    ]
```

tests/test_validators.py
mpsolano: Agreed, it shouldn't cause issues leaving it in there. But in terms of our automated testing guidelines, would this be considered a "relevant detail"? This is an optional suggestion of course so feel free to mark the conversation as resolved as I've already approved.

tests/test_validators.py
mpsolano: Ah understood, that makes a lot of sense, that we want to test a few combinations but not an exhaustive test. Thank you, do you mind adding a docstring to the test to make that intention clear to future maintainers?

src/gdcdictionary/validators.py
kulgan: You are right, it isn't extended by anything and if it extended, implementing those methods are optional as well so they are not made abstract. They are unimplemented because there are no default pre and post validation work that needs to happen while doing basic json validation.
In terms of what it can be used for - partial json validation can be a good example, in that scenario it will be extended and either the pre or post validation methods implemented to do custom things either before validation happens (where we inject values previously saved) or after validation (where we filter out validation failures that can be ignored). It is potentially useful, but nothing (within the center) is using it right now, but not sure if anyone out there might find it useful.

kulgan: Decided to remove it as it does not affect the current implementation, I like the idea of extending the validator but a different PR might be more appropriate for it

PR #1063, opened 2024-05-16
DEV-2693: add partials json schema validation

kulgan:
Implements validation for incomplete json schema used for patching previously submitted documents.


mpsolano (approved):
Looks good!

PR #1071, opened 2024-08-20
Okazaki - Molecular Test - P1

wwysoc2:


ebusygina (approved):
looks good to me

src/gdcdictionary/schemas/molecular_test.yaml
```diff
@@ -7333,6 +7331,12 @@ properties:
       - 76-100%
       - Unknown
       - Not Reported
+      - 0-25
+      - 26-50
+      - 51-75
+      - 76-100
+      - 1-4%
+      - 5-9%
```
wteouchicago: do these enums need to be organized (i.e. "sorted") differently, e.g. 1-4% and 5-9% should be before <10%? (or at least before 10-19%? 😵 ) or at least before Unknown and Not Reported?

wwysoc2: Oh yes, 1-4, 5-9 should be above Unknown and Not Reported.  I think the other four that were added should be deprecated.  I'll fix.

wteouchicago (approved):

stilesj-uchicago (approved):
DICT-523 phase 1 looks good to me

PR #1077, opened 2024-08-28
Okazaki - TERMS - DICT-510 

bullenc:
in summary:
- we are fixing typos for a few terms descriptions
- we are updating a few property terms with new NCIt codes/descriptions
- we are adding new properties in this release and thus will need to add new terms entries for these new properties

### DIFF
--- a/src/gdcdictionary/schemas/_terms.yaml
+++ b/src/gdcdictionary/schemas/_terms.yaml
@@ -1213,7 +1213,7 @@ comorbidities:
   common:
     description: >-
       The text term used to describe a patient with one or more diseases or conditions,
-      which coexists with the patient's malignant disease at the same time.
+      which coexist with the patient's malignant disease at the same time.
     termDef:
       term: >-
         Index Date To Comorbidity Day Count
--- a/src/gdcdictionary/schemas/_terms.yaml
+++ b/src/gdcdictionary/schemas/_terms.yaml
@@ -175,7 +175,7 @@ aids_risk_factors:
   common:
     description: >-
       The text term used to describe a risk factor of the acquired immunodeficiency
-      syndrome (AIDS) that the patient either had at time time of the study or experienced
+      syndrome (AIDS) that the patient either had at the time of the study or experienced
       in the past.
     termDef:
       term: >-
@@ -783,7 +783,7 @@ bmi:
 body_surface_area:
   common:
     description: >-
-      Numeric value used to represent the 2-dimensional extent of the body surface
+      Numeric value used to represent the two-dimensional extent of the body surface
       relating height to weight.
     termDef:
       term: >-
@@ -1212,8 +1212,8 @@ columnar_mucosa_present:
 comorbidities:
   common:
     description: >-
-      The text term used to describe a comorbidity disease, which coexists with the
-      patient's malignant disease.
+      The text term used to describe a patient with one or more diseases or conditions,
+      which coexists with the patient's malignant disease at the same time.
     termDef:
       term: >-
         Index Date To Comorbidity Day Count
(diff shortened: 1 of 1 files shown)


src/gdcdictionary/schemas/_terms.yaml
```diff
@@ -7700,7 +7718,92 @@ hormonal_replacement_therapy_status:
...
+      cde_version: 24.07e
+      term_url: "https://ncit.nci.nih.gov/ncitbrowser/ConceptReport.jsp?dictionary=NCI_Thesaurus&ns=ncit&code=C206479"
+      term_id: C206479
+      term_version: 24.07e
+
+population_group:
+  common:
+    description: >-
+      A group of individuals united by a common factor (e.g., geographic location,
+      ethnicity, disease, age, gender)
```
wteouchicago: do we need to have a period at the end of the desc? (it's missing in the ticket too.)

wteouchicago:
- I didn't see the change for "…described by the Center for Disease Control." for cdc_hiv_risk_factors
- undescended_testis_corrected_age's description and term were already present (not sure if we need to update the ticket). ditto size_extraocular_nodule's term and description.

ebusygina:
I see a small difference: the property name is `genome_doubling` in the code, and in the task `genome_doublings`.

src/gdcdictionary/schemas/_terms.yaml
bullenc: Great catch, added 👍

bullenc:
> * I didn't see the change for "…described by the Center for Disease Control." for cdc_hiv_risk_factors
> 
>     * undescended_testis_corrected_age's description and term were already present (not sure if we need to update the ticket). ditto size_extraocular_nodule's term and description.

Fixed ticket and yaml, should read 'Centers for Disease Control'.

Removed undescended_testis_corrected_age from ticket, you are correct, a term already existed for it so no need to be added to ticket as a new addition

bullenc:
> I see a small difference: the property name is `genome_doubling` in the code, and in the task `genome_doublings`.

fixed ticket 👍 should read `genome_doubling`

wteouchicago (approved):

ebusygina:
undescended_testis_corrected_age: in the code, we added NCIt code to it, probably we need to show it in the task


size_extraocular_nodule: in the code, we added NCIt code to it and didn't add its description and terms. probably, we need to delete its <new description> and <new term> from the task

src/gdcdictionary/schemas/_terms.yaml
```diff
@@ -1212,8 +1212,8 @@ columnar_mucosa_present:
 comorbidities:
   common:
     description: >-
-      The text term used to describe a comorbidity disease, which coexists with the
-      patient's malignant disease.
+      The text term used to describe a patient with one or more diseases or conditions,
+      which coexists with the patient's malignant disease at the same time.
```
wwysoc2: I seriously have no idea if "coexists" or "coexist" should be used here.

wwysoc2 (approved):
Looks good, had one comment about grammar.

src/gdcdictionary/schemas/_terms.yaml
wteouchicago: > I seriously have no idea if "coexists" or "coexist" should be used here.

me too. seems like "one or more" can take the singular or plural, but then the "which" can refer to either, so...

bullenc: Updated to "coexist", I think it flows better and makes more grammatical sense given diseases/conditions are plural

bullenc:
> undescended_testis_corrected_age: in the code, we added NCIt code to it, probably we need to show it in the task
> 
> size_extraocular_nodule: in the code, we added NCIt code to it and didn't add its description and terms. probably, we need to delete its and from the task

updated ticket to reflect this

ebusygina (approved):

stilesj-uchicago (approved):
Everything looks correct according to ticket DICT-510. My nit is that the terms are not in alphabetical order but that is not required.

PR #1078, opened 2024-08-29
okazaki p2 OCA updates

bullenc:


wwysoc2 (approved):
👍 lgtm!

ebusygina (approved):

ProfOak (commented):
Do we not need to have Parkinson's and dementia in the risk factor section?

stilesj-uchicago (approved):
Phase 2 looks good.

bullenc:
> Do we not need to have Parkinson's and dementia in the risk factor section?

IMS did not specify these as additions to risk_factors, and they just did a pretty significant comorbidity vs risk_factor assessment recently, so I am guessing for the disease in question (uveal melanoma) that these are not risk factors per se but tend to be mainly comorbidities when the disease shows up (which sounds like in older folks).

ProfOak (approved):

wteouchicago (approved):

PR #1080, opened 2024-08-29
okazaki p2 pathology detail updates

bullenc:


src/gdcdictionary/schemas/pathology_detail.yaml
```diff
@@ -1012,6 +1057,60 @@ properties:
...
+      - After Chemotherapy
+      - After Study Enrollment
+      - Childhood
+      - End of Consolidation Therapy
+      - End of Treatment Course
+      - End of Treatment Course 1
+      - End of Treatment Course 2
+      - First Complete Response
+      - First Treatment
+      - Follow-up
```
wwysoc2, line 1088: Should this be "Follow-Up"?  We can keep it lowercase if the other nodes have it like this.

wwysoc2 (approved):
Approved, but look at that one "Follow-up" value.  I'll leave the course of action up to you.

src/gdcdictionary/schemas/pathology_detail.yaml
bullenc: Yeah the letter case was set on the FollowUp node funny enough, which was then inherited by OtherClinicalAttribute. I think for sake of consistency we should keep as is and for other nodes, and then we can deprecate+migrate in the future.

ProfOak (approved):

stilesj-uchicago (approved):
LGTM

wteouchicago (approved):

PR #1082, opened 2024-08-29
okazaki p2 follow_up updates

bullenc:


src/gdcdictionary/schemas/follow_up.yaml
```diff
@@ -1595,8 +1595,11 @@ properties:
     $ref:
```
wwysoc2, line 1595: Verified that this is synced with evidence_of_recurrence_type

wwysoc2 (approved):
lgtm! 👍

ebusygina (approved):

ProfOak (approved):

src/gdcdictionary/schemas/follow_up.yaml
wteouchicago: double verified.

wteouchicago (approved):

PR #1087, opened 2025-06-10
TERMS - TURING - DICT-561

wwysoc2:
https://gdc-ctds.atlassian.net/browse/DICT-561

### DIFF
--- a/src/gdcdictionary/schemas/_terms.yaml
+++ b/src/gdcdictionary/schemas/_terms.yaml
@@ -7864,7 +7864,7 @@ tmb_exonic:
   common:
     description: >-
       Tumor Mutational Burden (TMB), reported in mutations per
-      megabase (mut/Mb), calculated within the intersect of targeted capture
+      megabase (mut/Mb), calculated within the intersect of the targeted capture
       region and exonic region.
     termDef:
       term: >-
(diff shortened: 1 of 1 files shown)


wwysoc2:
The two most important things to verify:  

1. That the fields in DICT-561 are represented accurately here. 
2. That the new links work and lead somewhere that makes sense to the field.

src/gdcdictionary/schemas/_terms.yaml
```diff
@@ -7807,3 +7807,112 @@ pretreatment:
...
+      cde_id:
+      cde_version:
+      term_url:
+
+tmb_nonsynonymous_exonic:
+  common:
+    description: >-
+      Tumor Mutational Burden (TMB) of nonsynonymous mutations only, reported
+      in mutations per megabase (mut/Mb), calculated within the intersect of
+      targeted capture region. and exonic region.
```
clsackett: Are there supposed to be two periods at the end of the description? I see that that matches what's in the JIRA ticket, but I'm wondering if it's a typo? "targeted capture region, and exonic region."? Or just "targeted capture region and exonic region."?

clsackett (commented):
One questionable typo, otherwise LGTM

src/gdcdictionary/schemas/_terms.yaml
wwysoc2: Good catch!  Removed period and added a few "the"

clsackett (approved):
LGTM!

kulgan (approved):

ebusygina:
LGTM! My only comment is that we might want the properties in alphabetical order in the _terms file

ebusygina (approved):

PR #1088, opened 2025-06-11
Turing updates to family_history node

clsackett:
Turing updates to family_history node by adding relationship_sex_at_birth and deprecating relationship_gender in support of [DICT-548](https://gdc-ctds.atlassian.net/browse/DICT-548)

### DIFF
--- a/src/gdcdictionary/schemas/family_history.yaml
+++ b/src/gdcdictionary/schemas/family_history.yaml
@@ -52,6 +52,32 @@ properties:
     maximum: 89
     minimum: 0
 
+  relationship_gender:
+    $ref:
+      - "_terms.yaml#/relationship_gender/common"
+    enum:
+      - female
+      - male
+      - unspecified
+      - unknown
+      - not reported
+    enumDef:
+      female:
+        $ref:
+          - "_terms_enum.yaml#/female/common"
+      male:
+        $ref:
+          - "_terms_enum.yaml#/male/common"
+      unspecified:
+        $ref:
+          - "_terms_enum.yaml#/unspecified/common"
+      unknown:
+        $ref:
+          - "_terms_enum.yaml#/unknown/common"
+      not reported:
+        $ref:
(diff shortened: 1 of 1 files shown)


src/gdcdictionary/schemas/family_history.yaml
```diff
@@ -49,32 +49,6 @@ properties:
     maximum: 89
     minimum: 0
 
-  relationship_gender:
```
wwysoc2: Add the relationship_gender field back since it should only be deprecated and not deleted yet.   I think you just need to add it to the list of deprecated fields near the top of the file.

wwysoc2 (changes requested):
I'm not sure if there is a "deprecated" section in the family history node yet, but create one if there is.  The diagnosis node will definitely have one.

src/gdcdictionary/schemas/family_history.yaml
clsackett: Added it back under a new deprecated section

wteouchicago: I think you'll need to undelete the lines that were deleted for this property. 

e.g. see `pregnancy_count` in https://github.com/NCI-GDC/gdcdictionary/blob/develop/src/gdcdictionary/schemas/other_clinical_attribute.yaml

wwysoc2: Also add back the deleted `relationship_gender` field that was deleted in the first commit.  Technically the field will still exist until we can migrate/remove from portal/etc.

clsackett: Made another commit to add relationship_gender back

wteouchicago (approved):

ebusygina (approved):

wwysoc2 (approved):
👍 lgtm!

stilesj-uchicago (approved):

PR #1091, opened 2025-06-11
DICT-551 updates to diagnosis node

clsackett:
DICT-551 updates to diagnosis node in support of [DICT-551](https://gdc-ctds.atlassian.net/browse/DICT-551)


wteouchicago (approved):
Just noting in case we want to remove entries from the _terms file -- I didn't see the following properties in other nodes, so we may not need their definitions anymore:
  - metastasis_at_diagnosis_site
  - micropapillary_features
  - papillary_renal_cell_type
  - primary_disease

Worth double-checking my list in case search isn't cooperating.

ebusygina (approved):

stilesj-uchicago (approved):

PR #1093, opened 2025-06-11
DICT-550 - Turing Analyte Changes

wwysoc2:
https://gdc-ctds.atlassian.net/browse/DICT-550


wteouchicago (approved):

zhangm-02 (approved):

ebusygina (approved):
just wanted to check that `dna_integrity_number` should have `minimum: 0`. if so, maybe we should update it in the task then

clsackett (approved):

PR #1096, opened 2025-06-11
Dict 556 updates to sample node

clsackett:
[DICT-556](https://gdc-ctds.atlassian.net/browse/DICT-556) updates to sample node


wteouchicago (approved):

zhangm-02 (approved):

ebusygina (changes requested):
1. probably we should also remove Brain Stem, Common Duct, and Indeterminant from the list of deprecated enums

2. also, Indeterminant has a definition here: https://github.com/NCI-GDC/gdcdictionary/blob/9e9ba6ffcfb592c40eda2dd730345dd908d9dce9/src/gdcdictionary/schemas/sample.yaml#L721

3. we have "method_of_sample_procurement": "Indeterminate" in the example file, maybe we want to update it as well

ebusygina (approved):

wwysoc2 (approved):
lgtm!  👍

stilesj-uchicago (approved):

PR #1101, opened 2025-06-12
DICT-529 updates to pathology_detail node

clsackett:
[DICT-529](https://gdc-ctds.atlassian.net/browse/DICT-529?linkSource=email)

### DIFF
--- a/src/gdcdictionary/schemas/pathology_detail.yaml
+++ b/src/gdcdictionary/schemas/pathology_detail.yaml
@@ -196,6 +196,15 @@ properties:
     type: number
     minimum: 0
 
+  breslow_thickness_category:
+    $ref:
+      - "_terms.yaml#/breslow_thickness_category/common"
+    enum:
+        - <=1 mm
+        - '>1.0-2.0 mm'
+        - '>2.0-4.0 mm'
+        - '>4.0 mm'
+
   circumferential_resection_margin:
     $ref:
       - "_terms.yaml#/circumferential_resection_margin/common"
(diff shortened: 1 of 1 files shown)


wteouchicago (approved):

ebusygina (approved):

stilesj-uchicago (approved):

src/gdcdictionary/schemas/pathology_detail.yaml
```diff
@@ -196,6 +196,15 @@ properties:
     type: number
     minimum: 0
 
+  breslow_thickness_category:
+    $ref:
+      - "_terms.yaml#/breslow_thickness_category/common"
+    enum:
+        - <=1 mm
```
wwysoc2, line 203: @clsackett   This enum isn't surrounded by the single-quotes.  Why is that?  I always forget what the rules are here.

clsackett: @wwysoc2 I was basing it on test_value_range on the [molecular_test](https://github.com/NCI-GDC/gdcdictionary/blob/develop/src/gdcdictionary/schemas/molecular_test.yaml) node. To me, it looks like if a value starts with '<', then it doesn't need quotes around it, but if it starts with '>' then the single quotes are required. Not sure why this is the case though!

wwysoc2: Good to know!

wwysoc2 (approved):
lgtm! 👍

PR #1149, opened 2025-06-24
DEV-3571: Standardize/fix usage of importlib.resources

BenX00:
Changes:
- Fix usages of `resources.files` to utilize standard interface for `abc.Traversable`
- Minor clean up of the core module.


stilesj-uchicago (approved):

src/gdcdictionary/core.py
```diff
@@ -66,45 +69,43 @@ def __init__(
...
 
     def load_schemas_from_dir(
-        self, directory: pathlib.Path
-    ) -> tuple[dict[str, Any], dict[str, Any]]:
+        self, directory: abc.Traversable
+    ) -> tuple[dict[str, dict], dict[str, ResolverPair]]:
         """Returns all yamls and resolvers of those yamls from dir"""
 
         schemas, resolvers = {}, {}
+        paths = (p for p in directory.iterdir() if p.name.endswith(".yaml"))
```
kulgan: I will recommend sticking with `.glob` (by either casting to a Path instance or just sticking with Path from the get go) instead of `iterdir`. It's fine tuned for things like this.

BenX00: Would you be able to point me to where this is documented? And more to the point would it matter at this scale?

BenX00: So ran this several times with timeit (just doing tuple(glob) & tuple(endswith) and I'm actually seeing endswith is quicker than the glob. I mean at the end of the day it's something like 0.00011 vs 0.000085 so really not a huge diff at all.

kulgan: A quick google search led to this stackoverflow [post](https://stackoverflow.com/questions/63001429/what-is-the-difference-between-pathlib-glob-and-iterdir). I think it matters for projects like plaster, gdcdatamodels, sheepdog and a couple qa projects that load the dictionary at some point since each extra millisecond adds to the startup time. The runtime diff you posted does help with decision making, but I will still stick to keeping the glob based implementation since filtering files in a directory is `glob`'s primary purpose (so I expect better performance  when compared to listing all the files and then doing the filtering yourself).
I think the more fundamental question is why the change was needed and I know its because `Traversable` does not have the `glob` method. For me, its worth re-evaluating when its best to use the `Traversable` and when to stick to simple `Path` objects. This particular use case looks like we are introducing changes solely to support the `Traversable` interface.

BenX00: At the end of the day, I don't think this is really a convo about what is right rather who. So happy to let you have a win on this one as it seems quite important to you.

kulgan: Not sure I get what you mean by not `really about what is right rather who`, in my experience the who is less important than what is right. Also, didn't know there's a race to win here, so I guess thanks for letting me win this one 👍🏾

kulgan (approved):

PR #1150, opened 2025-06-25
DICT-563 fertility field for Turing

wwysoc2:
https://gdc-ctds.atlassian.net/browse/DICT-563 - Specifically only the fertility field part.

### DIFF
--- a/src/gdcdictionary/schemas/other_clinical_attribute.yaml
+++ b/src/gdcdictionary/schemas/other_clinical_attribute.yaml
@@ -554,11 +554,11 @@ properties:
     $ref:
       - "_terms.yaml#/fertility_history/common"
     enum:
+      - Conceived > or = 1 child by assisted reproduction
+      - Conceived > or = 1 child by natural conception
+      - Conceived > or = 1 child by unspecified method
       - Did not achieve pregnancy following > or = 12 months of unprotected intercourse
       - Did not attempt to reproduce
-      - Fathered > or = 1 child by assisted reproduction
-      - Fathered > or = 1 child by natural conception
-      - Fathered > or = 1 child by unspecified method
       - Unknown
       - Not Reported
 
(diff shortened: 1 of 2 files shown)


clsackett (approved):

src/gdcdictionary/schemas/other_clinical_attribute.yaml
```diff
@@ -550,6 +550,18 @@ properties:
         $ref:
           - "_terms_enum.yaml#/not_reported/common"
 
+  fertility_history:
+    $ref:
+      - "_terms.yaml#/fertility_history/common"
+    enum:
+      - Conceived > or = 1 child by assisted reproduction
+      - Conceived > or = 1 child by natural conception
+      - Conceived > or = 1 child by unspecified method
```
wteouchicago: probably above my pay grade, but we use ">=" in other enums (though slightly different contexts) -- should we try to be consistent? and are we open to other phrasings, e.g. "1 or more", "1+", "at least 1"

wwysoc2: Good point!  My thought would be that the following would be acceptable: 

- Conceived 1+ child by assisted reproduction
- Conceived at least 1 child by assisted reproduction

@tlicht3 do you have any thoughts?  Are there guidelines that should be followed?

tlicht3: Yeah I think that's a good suggestion - without the symbols.

tlicht3:
@wwysoc2 this looks great to me :)

tlicht3 (approved):
@wwysoc2 this looks great!

stilesj-uchicago (approved):

src/gdcdictionary/schemas/other_clinical_attribute.yaml
```diff
@@ -550,6 +550,18 @@ properties:
         $ref:
           - "_terms_enum.yaml#/not_reported/common"
 
+  fertility_history:
+    $ref:
+      - "_terms.yaml#/fertility_history/common"
+    enum:
+      - Conceived 1 or more child by assisted reproduction
+      - Conceived 1 or more child by natural conception
+      - Conceived 1 or more child by unspecified method
```
wteouchicago: prob. too pedantic...but should "1 or more" take a singular or plural noun?

wteouchicago (approved):

PR #1157, opened 2025-07-03
margin_status updates to pathology_detail

clsackett:
The update to pathology_detail.margin_status was somehow missed before. Adding this in support of [DICT-529](https://gdc-ctds.atlassian.net/browse/DICT-529)

### DIFF
--- a/src/gdcdictionary/schemas/pathology_detail.yaml
+++ b/src/gdcdictionary/schemas/pathology_detail.yaml
@@ -701,6 +701,8 @@ properties:
       - Uninvolved
       - Unknown
       - Not Reported
+    deprecated_enum:
+      - Indeterminant
     enumDef:
       Indeterminate:
         $ref:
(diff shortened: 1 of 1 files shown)


wteouchicago (approved):

tlicht3 (commented):
@clsackett I think this is ok, but if someone is already using the incorrect value, will this have an impact on that? Not sure if that's ok or if you've already checked to see if that value exists in the data.

tlicht3 (approved):

src/gdcdictionary/schemas/pathology_detail.yaml
```diff
@@ -696,15 +696,17 @@ properties:
...
       - "_terms.yaml#/margin_status/common"
     enum:
-      - Indeterminant
+      - Indeterminate
       - Involved
       - Uninvolved
       - Unknown
       - Not Reported
+    deprecated_enum:
+      - Indeterminant
```
wteouchicago, line 706: i think a deprecated enum needs to appear in the enum list too? not 100% sure, but I see examples of that elsewhere.

and do we need to keep the enumdef for the deprecated one?

wteouchicago (approved):

stilesj-uchicago (approved):

PR #1160, opened 2025-09-03
DEV-3245: Added deprecated properties/enums explanation to README

ebusygina:
DEV-3245: Added deprecated properties/enums explanation to README

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,6 +64,26 @@ such as `submitter_id`, will allow any string value as a valid entry. Other nume
 can have maximum and minimum values to limit valid entries.  For examples of what a valid entry
 would look like, each node has a mock submission located in the `examples/valid/` directory.
 
+### Deprecated Properties and Enums
+
+The GDC Dictionary uses deprecated flags for properties and enums as part of the phased removal process. When a property or enum value is marked as deprecated:
+
+- It remains present in the dictionary, data models, API, and databases.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -64,6 +64,26 @@ such as `submitter_id`, will allow any string value as a valid entry. Other nume
...
+- `properties`: Lists all available properties for submission.
+- `deprecated`: Hides the property from the dictionary viewer, but it remains available for submission.
+
+**For enums**:
+
+- `enum`: Indicates a permissible value for submission.
+- `deprecated_enum`: Hides the value from the dictionary viewer, but it remains available for submission.
+
+This approach allows for a transition period before full removal, ensuring deprecated properties and enums are still accessible in system integrations while not displayed to dictionary users.
+
```
wwysoc2, line 86: I was going to recommend a bunch of "as long as the original enum/property is still present", but I think we should add something to the end instead like:

"Note:  deprecated properties or enums will only still be valid if they are also present in the permissible properties/enums section."

wwysoc2 (approved):
Added an optional recommendation.  It covers what I think is an edge case (value is present as deprecated, but already removed), but if one is around, that edge case will stick around for months.  Feel free to add or not.  The rest looks good!  👍

README.md
ebusygina: Thanks! Added

PR #1168, opened 2026-04-23
DICT-601: Updates to treatment node for WS release

wwysoc2:
https://gdc-ctds.atlassian.net/browse/DICT-601

### DIFF
--- a/src/gdcdictionary/schemas/treatment.yaml
+++ b/src/gdcdictionary/schemas/treatment.yaml
@@ -38,9 +38,6 @@ uniqueKeys:
   - [id]
   - [project_id, submitter_id]
 
-deprecated:
-  - therapeutic_level_achieved
-
 properties:
 
   $ref:
@@ -2191,6 +2188,7 @@ properties:
       - Dynemicin A
       - E2F1 Pathway Activator ARQ 171
       - EBNA-1 inhibitor VK-2019
+      - EBV CTLs
       - Echinomycin
       - Ecromeximab
       - Edatrexate
@@ -3758,6 +3756,7 @@ properties:
       - PEG-PEI-cholesterol Lipopolymer-encased IL-12 DNA Plasmid Vector GEN-1
       - PEG-Proline-Interferon Alfa-2b
       - Pegargiminase
+      - Pegasparagase
       - Pegaspargase
       - Pegdinetanib
       - Pegfilgrastim
@@ -4669,6 +4668,7 @@ properties:
       - Tocladesine
       - Tocotrienol
       - Tocotrienol-rich Fraction
+      - Tofacitinib
       - Tolebrutinib
       - Toll-like Receptor 7 Agonist DSP-0509
       - Tolnidamine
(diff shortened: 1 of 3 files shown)


src/gdcdictionary/schemas/treatment.yaml
```diff
@@ -7606,9 +7590,87 @@ properties:
...
-      Not Allowed To Collect:
-        $ref:
-          - "_terms_enum.yaml#/not_allowed_to_collect/common"
+
+  treatment_type_administered:
+    $ref:
+      - "_terms.yaml#/treatment_type_administered/common"
+    oneOf:
+      - type: "null"
+      - enum:
```
wwysoc2, line 7599: I would really like to point this list to the enums in treatment_type, but I'm not quite sure how to do that in the gdcdictionary framework.

clsackett (approved):

Inarus899 (approved):

mpsolano (approved):

src/gdcdictionary/schemas/treatment.yaml
```diff
@@ -4993,6 +4995,8 @@ properties:
       - Other
       - Unknown
       - Not Reported
+    deprecated_enum:
+      - Pegaspargase
```
wteouchicago, line 4999: I didn't see this in the ticket as something to deprecate. Are we trying to hide it from the dictionary viewer or something?

src/gdcdictionary/schemas/treatment.yaml
```diff
@@ -4993,6 +4995,8 @@ properties:
       - Other
       - Unknown
       - Not Reported
+    deprecated_enum:
```
ebusygina, line 4998: I don't see this deprecation in the DICT-601. should we update the ticket?

src/gdcdictionary/schemas/treatment.yaml
```diff
@@ -7606,9 +7590,87 @@ properties:
       Not Reported:
         $ref:
           - "_terms_enum.yaml#/not_reported/common"
-      Not Allowed To Collect:
-        $ref:
-          - "_terms_enum.yaml#/not_allowed_to_collect/common"
+
+  treatment_type_administered:
```
ebusygina, line 7594: Should we also add the `enumDef` section for this property, or is that not necessary? I’m not exactly sure if we use definitions from _terms_enum.yaml anywhere

ebusygina (approved):

src/gdcdictionary/schemas/treatment.yaml
wwysoc2: Exactly, we were told to keep this enum defs around, but no need to add to them any further.  Thank you for checking.

src/gdcdictionary/schemas/treatment.yaml
wwysoc2: Added to jira ticket, it was a misspelling and the correct value was added in this PR.

src/gdcdictionary/schemas/treatment.yaml
wwysoc2: Added to jira ticket, it was a misspelling and the correct value was added in this PR.

wteouchicago (approved):

PR #1170, opened 2026-04-23
DICT-605: Updates to molecular test node for WS release

wwysoc2:
https://gdc-ctds.atlassian.net/browse/DICT-605

### DIFF
--- a/src/gdcdictionary/schemas/molecular_test.yaml
+++ b/src/gdcdictionary/schemas/molecular_test.yaml
@@ -7312,10 +7312,6 @@ properties:
     $ref:
       - "_terms.yaml#/test_value_range/common"
     enum:
-      - 0-25
-      - 26-50
-      - 51-75
-      - 76-100
       - 1-4%
       - 5-9%
       - <10%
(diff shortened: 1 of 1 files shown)


clsackett (approved):

Inarus899 (approved):

src/gdcdictionary/schemas/molecular_test.yaml
```diff
@@ -7352,7 +7343,6 @@ properties:
       - Adulthood
       - After Chemotherapy
       - After Study Enrollment
-      - After Study Registration
```
mpsolano, line 7355: Looks like the wrong enum may have gotten removed here, I believe the intention was to remove `After Study Enrollment`

wteouchicago: I have the same comment as Mike does on whether the wrong enum was removed.

wwysoc2: Good catch!  Fixing now.

wteouchicago (approved):

mpsolano (approved):

PR #1173, opened 2026-04-29
DICT-599: Updates to pathology_detail node for WS release

wwysoc2:
https://gdc-ctds.atlassian.net/browse/DICT-599

### DIFF
--- a/src/gdcdictionary/schemas/pathology_detail.yaml
+++ b/src/gdcdictionary/schemas/pathology_detail.yaml
@@ -38,12 +38,6 @@ uniqueKeys:
   - [id]
   - [project_id, submitter_id]
 
-deprecated:
-  - epithelioid_cell_percent
-  - extrascleral_extension
-  - size_extraocular_nodule
-  - spindle_cell_percent
-
 properties:
 
   $ref:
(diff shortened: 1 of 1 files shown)


src/gdcdictionary/schemas/pathology_detail.yaml
```diff
@@ -396,12 +383,7 @@ properties:
       - <=5mm
       - '>5mm'
       - Unknown
-      - Not Reported
-
-  extrascleral_extension:
-    $ref:
-      - "_terms.yaml#/extrascleral_extension/common"
-    type: boolean
+      - Not Reporte
```
BenX00: Think there may have been a typo introduced.

```suggestion
      - Not Reported
```

src/gdcdictionary/schemas/pathology_detail.yaml
```diff
@@ -628,10 +604,13 @@ properties:
       - "_terms.yaml#/lymph_node_involvement/common"
     enum:
       - Indeterminant
+      - Indeterminate
```
BenX00, line 607: Do we need to add this to the enumDef section as well?

src/gdcdictionary/schemas/pathology_detail.yaml
wwysoc2: Good catch, fixed!

src/gdcdictionary/schemas/pathology_detail.yaml
wwysoc2: No need, but thanks for asking! We decided to stop using them, but a third party group is currently using them for a different project. We aren't removing them yet, but we aren't actively updating them.

ebusygina (approved):

wteouchicago (approved):

PR #1179, opened 2026-09-17
PEAR-2747: Add pull request template with pre-merge checklist

ebusygina:
PEAR-2747: Adds a pull request template to the repository so that new PRs are 
automatically populated with a pre-merge checklist.

- The template only takes effect once this is merged into the default 
  branch - templates on feature branches aren't picked up.
- After merging, opening any new PR should show the checklist in the 
  description field automatically.


pull_request_template.md
```diff
@@ -0,0 +1,6 @@
+## Description
+
+## Checklist
+
+- [ ] New properties (if any) follow the front-end [property type rules](https://gdc-ctds.atlassian.net/wiki/spaces/GDC/pages/1131610113) for Cohort Builder and Repository
```
wteouchicago: We should not force properties to follow the logic the FE has set up. We should be checking new properties against the FE logic and adjusting the FE logic if needed. 

If it's something that the User Services team had done in the past but didn't do for some property (e.g. including "year" in definition for age property), that's something we can point out.

wteouchicago (changes requested):

pull_request_template.md
wteouchicago: Maybe just rephrase it to be clearer? And state that it's the FE/Product team doing the review that should check this? Otherwise, people opening the PRs may think they are responsible for it.

wteouchicago (approved):
Wording works for me -- pending @wwysoc2 's approval, of course. I briefly wondered if we should put this in a different dir, but realized that the GFF repo does the same thing.

wwysoc2 (approved):
lgtm! 👍

